PR #31, opened 2017-02-12
Mention that only store/index.js receives the nuxtServerInit action

samtgarson:
Example issues: 
- https://github.com/nuxt/nuxt.js/issues/169#issuecomment-275759867
- https://github.com/nuxt/nuxt.js/issues/175


Atinux:
Thank you @samtgarson

PR #45, opened 2017-02-28
Add documentation to have ava transpile imported files

Austio:
Adds documentation to ava end to end tests to transpile imported files with babel.  

https://github.com/avajs/ava/blob/master/docs/recipes/babelrc.md

Note: This still does not have recommendation on how to transpile .vue files that are imported outside of nuxt page tests


Atinux:
Thank you @Austio

PR #46, opened 2017-02-28
Add a caveat to the Page data function docs

samtgarson:
To notify about issue https://github.com/nuxt/nuxt.js/issues/283


samtgarson:
Not sure why that security check is failing on this...

Atinux:
Since we fixed the mixins in the next release, we don't need to set this alert anymore :)

samtgarson:
👌 😉

PR #65, opened 2017-03-30
Wording

cretueusebiu:
the navigation will ~~fill like~~ _feel_ instant ~~will~~ _while_ keeping the code splitted


alexchopin:
Thanks 👍

PR #67, opened 2017-04-01
Reflect the update of Japanese translation repository name

inouetakuya:
I have reflected the update of Japanese translation repository name.


Atinux:
🇯🇵

PR #68, opened 2017-04-01
Add the documentation about CNAME file for GitHub Pages

inouetakuya:
I saw the question about handling the CNAME file for GitHub Pages yesterday in Japanese Vue.js community. I think it is useful to add the documentation about it.


Atinux:
Nice catch @inouetakuya

PR #70, opened 2017-04-02
Reflect the update of #68 to Japanese translation

inouetakuya:
I have moved the documentation about CNAME file out of command line deployment because CNAME file has little relevance with command.

Then I have reflect the update of #68 to Japanese translation.


alexchopin:
👍

PR #85, opened 2017-04-18
Reflect "update middleware alias · nuxt/docs@be36811" to Japanese translation

inouetakuya:
I have reflected "update middleware alias · nuxt/docs@be36811" to Japanese translation.


willin:
not add yet in chinese

PR #99, opened 2017-04-25
Dynamic transition depending on the context

OlzhasAlexandrov:


Atinux:
Hi @OlzhasAlexandrov 

Actually, the transition method does not receive the context as a parameter, see https://nuxtjs.org/api/pages-transition

Atinux:
But it may be useful to give it as 3rd argument I think.

OlzhasAlexandrov:
@Atinux Have you tried it? It works:
https://glitch.com/edit/#!/join/7dc4ddfd-5332-4e49-8f3f-c7f08af53c1b

Atinux:
@OlzhasAlexandrov it's not the context but 2 arguments, the page where it's navigating to and the page where he's leaving.

PR #145, opened 2017-06-11
Update store examples to reflect v1 changes

acidjazz:
So that others don't stumble across this as well, helps out [#754](https://github.com/nuxt/nuxt.js/issues/754) and fixes [#787](https://github.com/nuxt/nuxt.js/issues/787)


johnRivs:
I still see the old example on the website.

cho-is:
I still see the old example too.

PR #164, opened 2017-06-30
Provides more information to avoid nuxt/nuxt.js#1003

Tabrizian:
Hello there :)
Loading css files from current directory requires ```node-sass``` and ```sass-loader``` but it's not stated in the documentation. Same issue here also: nuxt/nuxt.js#1003

Hope it helps


pi0:
What's your idea adding some condition like "if you want to use sass.." ?

Tabrizian:
Sure I'll fix it Pooya :)

Tabrizian:
I've used better sentencing. @pi0

Tabrizian:
@alexchopin It would be great if you can take a look at this PR. :))

PR #167, opened 2017-07-06
Update configuration-css.md

escapedcat:
added the deps info refering to this issue: https://github.com/nuxt/nuxt.js/issues/1003#issuecomment-312304831

### DIFF
--- a/en/api/configuration-css.md
+++ b/en/api/configuration-css.md
@@ -30,3 +30,9 @@ module.exports = {
   ]
 }
 ```
+
+## Dependencies
+For certain files you need extra dependencies.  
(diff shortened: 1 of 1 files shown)


Atinux:
Hi @escapedcat 

Please consider writing the solution instead of linking to an issue, it's better for the user experience :)

escapedcat:
I did add the solution (the info from the issue):
```
If you use `sass`/`scss` make sure to install these via npm:
- node-sass
- sass-loader
```
...plus the link to the issue as a reference.  

The issue commeny says:
> Iman, please ensure sass  dependencies are installed (node-sass and sass-loader) Also there is no need to directly specifying lang if using nuxt 1.x as it is automatically loaded by webpack :)

I can remove the link but I personally like to refer to backgrounds of things if people want to reas up on this. If you don't want the link and solution only I can remove the link. Let me know.

PR #169, opened 2017-07-13
add production flag for now deployments

crswll:
I wasn't able to deploy to `now` without this flag so here it goes.


Atinux:
Hi @crswll 

I deployed on now from the beginning without this flag. Mainly for `nuxt build` since we add it into webpack configuration directly.

It was probably a bug from now.sh

PR #172, opened 2017-07-18
Add note for VS Code specific ESLint settings

homerjam:


Atinux:
Hi @homerjam 

Sorry since we will have to add a section for every editor directly and depends on them, I believe having eslint plugin inside Visual Studio Code does the work perfectly :)

homerjam:
Sorry, not sure I follow. Are you saying eslint does this automatically now?

The reason I added the note was to reduce friction when starting to use Vue - it took me a few searches to find the answer on stack overflow. Perhaps somewhere else would be more appropriate but IDE specific info is pretty useful to have in the docs imho.

PR #188, opened 2017-08-03
Fix documentation about google analytics, according to rc3

nyl-auster:
use context.app.router instead of ~router
mention existing module
export plugin as a default export


Atinux:
Merci :)

PR #197, opened 2017-08-14
[typo] fix docs on vuex-store.md

jameswragg:
Correct docs as mentioned in #757

### DIFF
--- a/en/guide/vuex-store.md
+++ b/en/guide/vuex-store.md
@@ -20,7 +20,7 @@ Nuxt.js lets you have **2 modes of store**, choose the one you prefer:
 
 ## Classic mode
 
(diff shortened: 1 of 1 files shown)


en/guide/vuex-store.md
```diff
@@ -20,24 +20,25 @@ Nuxt.js lets you have **2 modes of store**, choose the one you prefer:
...
 
-const store = () => new Vuex.Store({
-
-  state: {
-    counter: 0
-  },
-  mutations: {
-    increment (state) {
-      state.counter++
+const createStore = () => {
```
pi0: What's wrong with simply returning new Vuex Store using arrow shortcuts?

Atinux: I updated it since it does not have to be named 'createStore' but indeed the name is more explicit.

PR #199, opened 2017-08-16
Update configuration-env.md

uptownhr:
explains why process.env == {}


pi0:
related to nuxt/nuxt.js#1386, nuxt/nuxt.js#1323

PR #203, opened 2017-08-21
Create context.md

uptownhr:
The context is a crucial piece to understanding and working with nuxt. It's been my experiencing that it was hard to get to the documentation page describing the context as it does not have it's own page and available within the asyncData page.

Also added `app` as one of the objects available in the `context`

### DIFF
--- a/en/api/context.md
+++ b/en/api/context.md
@@ -1,0 +1,25 @@
+---
+title: "API: The Context"
+description: The `context` provides additional objects/params from Nuxt not traditionally available to Vue components. The `context` is available in special nuxt lifecycle areas like `asyncData`, `plugins`, 'middlewares', 'modules', and 'store/nuxtServerInit`. 
+---
+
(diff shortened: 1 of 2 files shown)


uptownhr:
Wasn't sure how I can create a link from the navigation for this. Can someone point in the right direction on the steps that is required to do so?

pi0:
Nice idea @uptownhr . Please check out `menu.json` file for menu :)

uptownhr:
@pi0 which project is the `menu.json` located in?

pi0:
@uptownhr There is one menu per section :) Here: https://github.com/nuxt/docs/blob/master/en/api/menu.json

uptownhr:
Let me know if we should remove the context keys tables from asyncData so we don't have to worry about updating in both places.

Atinux:
Thank you @uptownhr :)

PR #210, opened 2017-08-31
Fix Japanese format the same as English

inouetakuya:
I have fixed Japanese format the same as English so that I fix conflicts on GitLocalize.

https://gitlocalize.com/repo/100

I will translate with GitLocalize later.

CC: @sotayamashita


sotayamashita:
:shipit:

PR #217, opened 2017-09-08
Update Testing Documentation

haffla:
Hey,

this updates the end-to-end testing doc with the current Nuxt API. 

I also updated the comments in the code, but only for the english version as I don't speak any of the other languages. So I would need help there.

I noticed that in the Russian guide the docs are in `unit-testing.md` instead of `development-tools` resulting in a 404 [here](https://ru.nuxtjs.org/guide/development-tools/). This is certainly not unit testing is it? I can't read Russian though, so I don't know what it says in that file. So again, help needed!

### DIFF
--- a/en/guide/development-tools.md
+++ b/en/guide/development-tools.md
@@ -64,15 +64,14 @@ We add our test file `test/index.test.js`:
 
 ```js
 import test from 'ava'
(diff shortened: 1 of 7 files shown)


Atinux:
Wow! Thank you @haffla :clap:

PR #224, opened 2017-09-09
Add clearer nuxt-with-express example

trusktr:
For someone with only basic Express knowledge and trying Nuxt for the first time with only basic Vue experience, and having an existing Express app, this example is more clear than the other current two examples.


Atinux:
@trusktr actually we keep the link to the official repos, even is they might be a bit difficult at first sight

PR #225, opened 2017-09-09
listen after done building.

trusktr:
This seems to make more sense.

If this is the first time running the app, not in dev mode, and `.nuxt/` does not yet exists, will the app be able to function if it is not built? Is there a non-dev-mode build that should be first called?


Atinux:
The app will be working even if the build is running, but at least your example is easier to understand :)

PR #271, opened 2017-10-03
Update plugins documentation to include inject

samtgarson:
Example used: https://github.com/nuxt/nuxt.js/blob/b48b2c5e6d0834b7f760ec5abd7a8170da9a1f7c/examples/with-cookies/plugins/cookies.js#L15-L19


Atinux:
Thank you @samtgarson :)

PR #295, opened 2017-10-12
[Doc FR]: Credits for French translation

Haeresis:
Old repository is not relevant anymore @Atinux.

That PR following the @rspt proposition here : https://github.com/nuxt/docs/issues/294 (I let @rspt check if all is correct for him).


rspt:
Perfect for me :)

PR #302, opened 2017-10-16
[Doc FR]: `asyncData` in french

Haeresis:
Hi @Atinux, because the name file of asyncData is `index.htm` we have forgotten to add it amongs translated pages :)

This is the correction of that !


Atinux:
Awesome!!!

:confetti_ball:

PR #309, opened 2017-10-18
Update translation section on README

sotayamashita:
I have removed original text. If it is not good, please correct me. Thanks in advance.

cc:/ @gbezyuk


gbezyuk:
Hi there and thanks for the CC.

PR #313, opened 2017-10-24
fix: inject app in plugin

clarkdo:


Atinux:
Thanks, but actually vue-i18n should not be used with inject. Could you remove the inject code?

I believe we should add another part for inject since it’s pretty awesome :)

clarkdo:
@Atinux Done :)

Atinux:
Thank you @clarkdo :)

PR #318, opened 2017-10-27
Remove deprecated `cache` documentation

JesusCrow:
Remove `cache` listing from configuration documentation.
This feature was deprecated, and corresponding documentation, that this leads to, is gone.


pi0:
Thanks!

PR #319, opened 2017-10-27
Delete cached components example

JesusCrow:
This feature was deprecated, and corresponding documentation, that this leads to, is gone.
This example is currently misleading and should be deleted.
It would probably be a great idea to provide a replacement entry in FAQ page. Leading to [@nuxtjs/component-cache](https://github.com/nuxt-community/modules/tree/master/packages/component-cache)


pi0:
@JesusCrow PRs always welcome!

PR #320, opened 2017-10-27
Create cached-components.md

JesusCrow:
Created information about cached components in docs FAQ part.
While writing it, started to consider if proposed [module](https://github.com/nuxt-community/modules/tree/master/packages/component-cache) is official and proper solution.
In it's repository I couldn't find anything about option to pick, which components should receive caching.
Waiting for input from @pi0.

### DIFF
--- a/en/faq/cached-components.md
+++ b/en/faq/cached-components.md
@@ -1,0 +1,40 @@
+---
+title: Caching Components
+description: How to cache components?
+---
+
+# How to cache Vue components?
+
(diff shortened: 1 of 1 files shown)


pi0:
Thanks! I think we should add an entry into [menu.json](https://github.com/nuxt/docs/blob/master/en/faq/menu.json) too.

JesusCrow:
#321 should complete it 😃

PR #326, opened 2017-11-01
Fix hotMiddleware menu link

paulgv:
The _hotMiddleware_ link in the API menu does not point to the right anchor in EN and FR.


Atinux:
Thank you @paulgv

PR #327, opened 2017-11-03
a jwt authentication with nuxt.js example

ahadyekta:
An example about using JWT authentication with nuxt (with external api webservice)


ahadyekta:
hey @alexchopin , Is there any problem with this example ? I`ll be glad if there is some advice to improve it and merge it to the docs.

husayt:
Seems all cool. Thanks

PR #347, opened 2017-12-12
Added documentation for exporting vuex state, mutations, actions as object 

talentedunicorn:
This is an alternative way to export vuex state, mutations, actions in nuxtjs apart from exporting them as functions. It's similar to the modular function. [See example](https://github.com/nuxt/nuxt.js/tree/dev/examples/vuex-store-modules/store) but without requiring an `index.js` root file.

Unicorn brush

### DIFF
--- a/en/guide/vuex-store.md
+++ b/en/guide/vuex-store.md
@@ -164,6 +164,24 @@ export default {
 }
 </style>
 ```
+> The module method also works for top-level definitions without implementing a sub-directory in the `store` directory
+
+Example for state; you create a file `store/state.js` and add the following
+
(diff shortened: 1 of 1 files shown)


husayt:
don't you still need to add an `index.js` ?

talentedunicorn:
@husayt no you won't need `index.js` 
You can export your state as `state.js` and `mutations.js` 
See this [commit](https://github.com/nuxt/docs/pull/347/commits/7098538766e9b2270371cc4ef689d81575c7c9f8)

Atinux:
🙌🏻

PR #353, opened 2017-12-19
layout/view.vue now ignores the layout attribute

mvgrimes:
Per https://github.com/nuxt/nuxt.js/issues/1566 it seems layout/view.vue
is now a standalone page that cannot be composed with other layouts due
to conflicts.

I wasn't sure exactly how to handle the translations, so changed the source code
and inserted the comment in English so it doesn't get lost.

### DIFF
--- a/en/guide/views.md
+++ b/en/guide/views.md
@@ -86,7 +86,7 @@ export default {
 </script>
 ```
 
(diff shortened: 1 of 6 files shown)


Atinux:
Actually with v1, `layout` propety in `layouts/error.vue` will work :)

PR #362, opened 2018-01-04
Documented Fetch Use With Vuex

Mikeysax:
Added missing edge case to `fetch` usage with Vuex `dispatch`. If this is not done when using Vuex with Nuxt, then the Component `fetch` method will not trigger on page refresh.

If I am in any way incorrect, please correct me in possible alternatives. In my own experience, using `fetch` will not work otherwise.

Is this intentional functionality? If so it should be documented. Otherwise, if this is not intentional, then there is something much deeper that needs to be fixed.

This was also discussed previously: https://github.com/nuxt/nuxt.js/issues/1852

### DIFF
--- a/en/api/pages-fetch.md
+++ b/en/api/pages-fetch.md
@@ -48,3 +48,24 @@ export default {
 }
 </script>
 ```
+
+## Using Vuex
+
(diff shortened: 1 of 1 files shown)


Atinux:
Thanks @Mikeysax 

I updated a bit to explain why we need to wait.

Mikeysax:
Awesome 👍

PR #366, opened 2018-01-05
Update pages-head.md

samAZI:
replace key by the one use in vue-head called `vmid`


Atinux:
Hi @samAZI 

Actually in Nuxt.js, the key is `hid`, see https://github.com/nuxt/nuxt.js/blob/dev/lib/app/index.js#L35

PR #375, opened 2018-01-14
Fix some syntax errors

tonetlds:
Line 70:17
- Syntax Error: Unexpected token, expected (

Line 72:4
- 'store' is not defined


pi0:
Thanks :)

PR #383, opened 2018-01-18
Add <no-ssr> component document (Only /en)

potato4d:
## About

Nuxt include a `<no-ssr>` component, but documentation about it does not exist.
So, I added to documentation it.

## Remarks

The quoting source for `<no-ssr>` seems to be `egoist/vue-no-ssr`.
Should I add reference to it? Just like `<nuxt-link>` 's document.

### DIFF
--- a/en/api/components-no-ssr.md
+++ b/en/api/components-no-ssr.md
@@ -1,0 +1,23 @@
+---
+title: "API: The <no-ssr> Component"
+description: Skip component rendering on server side rendering, and rendering placeholder text.
+---
+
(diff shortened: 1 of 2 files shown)


pi0:
@potato4d Thanks for this PR. Yes, i think it would be better to add credits back to the original author. Also [egoist/vue-no-ssr](https://github.com/egoist/vue-no-ssr) would be used as a good reference to read more.

potato4d:
@pi0

Thanks for your comment!
I am also frequently helped by vue-no-ssr myself 🤗 .
Credit added!

PR #401, opened 2018-01-30
Updated recommended eslint config

husayt:
Given setup is quite outdated and am not sure if even works. Eslint works much better with `eslint-plugin-vue` and `prettier`. Not only this allows to report ,most of the issues, but also to fix them

This setup also nicely plays with code editors and VS code in particular.

### DIFF
--- a/es/guide/development-tools.md
+++ b/es/guide/development-tools.md
@@ -115,10 +115,11 @@ jsdom has some limitations because it does not use a browser. However, it will c
 
 ## ESLint && Prettier
 
-> ESLint is a great tool to keep your code clean
-> Prettier is a very popular formatter 
+> [ESLint](http://eslint.org) is a great tool to keep your code clean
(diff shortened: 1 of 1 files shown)


Atinux:
Thank you :)

husayt:
No, Thank you for enabling all this.

Tooling is the key though.

I want to add a section on editor integration and storybook. Also plan to
add storybook example. Let's put best practices here so that people can
benefit from most of the nuxt's features.

Atinux:
Would be awesome :)

husayt:
Here is the example project on nuxt+storybook+vuetify+chartjs
https://github.com/nuxt/nuxt.js/pull/2773
I did add quite a few usecases, just to demonstrate various approaches

husayt:
seems i added this to ES folder instead of EN will do another commit now

PR #434, opened 2018-02-06
[Doc FR]: Add new translation to API

Haeresis:
Hi @Atinux ! Some new translation for API part !


Atinux:
Super merci @Haeresis :)

PR #527, opened 2018-03-20
fix spell miss: 'generete'

sKawashima:
line 12


Atinux:
Thank you @sKawashima

PR #528, opened 2018-03-22
add render:errorMiddleware info

DiederikvandenB:
See https://github.com/nuxt/nuxt.js/issues/1688#issuecomment-374883057


Atinux:
🙌

PR #548, opened 2018-04-04
Fix typo

ruedap:
http://coffeescript.org/


ruedap:
oops, duplicated with #547

PR #549, opened 2018-04-04
feat: complete docs for genreation hooks

motia:
Hi, I filled the empty cells for the generation hook arguments and their description.


Atinux:
Thanks!

PR #556, opened 2018-04-10
Update Vuex Classic mode example to working condition

jpoechill:
The Classic mode example for using Vuex with Nuxt does not work above unless modified as shown. 

'this.$store.state' is not accessible unless accessed by a getter method.


jpoechill:
Sincerest apologies. Some reason, this was not working unless with fix, but now it seems to be fine. 

Please ignore, or possibly delete.

PR #558, opened 2018-04-13
fix typo

ljwrer:


husayt:
Although, 1000 is not 1 KB either, but 1K

PR #575, opened 2018-04-27
[Doc FR + EN]: Update documentation with last changes

Haeresis:
Hi @Atinux !

Just to be sure, the `## Tapable plugins` is not replaced by `## Hooks` in the only file with `Plugin` is replaced by Hook (en/api/internals-builder.md)

But all file into  `## Tapable plugins` was effectively replaced by `## Hooks`, the `Plugin` column is not change by Hook !

So I change all « Tapable plugins » by « Hooks » and all « Plugin » by « Hook ».

It's normally ok, and it's done in french

Ready to merge ?!


Atinux:
Wow! Thank you so much @Haeresis

PR #599, opened 2018-05-18
misguiding info about implementation for `script` into head

mesqueeb:
I spent a few hours trying to debug as I thought I had to refer to the vue-meta implementation method of `script` which is very different from the Nuxt way.

### DIFF
--- a/en/api/pages-head.md
+++ b/en/api/pages-head.md
@@ -5,7 +5,7 @@ description: Nuxt.js uses vue-meta to update the headers and HTML attributes of 
 
 # The head Method
 
-> Nuxt.js uses [vue-meta](https://github.com/declandewet/vue-meta) to update the `headers` and `html attributes` of your application.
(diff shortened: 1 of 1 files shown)


Atinux:
Could you explain a bit more about this @mesqueeb ?

Because Nuxt use vue-meta for everything inside the `head` key

mesqueeb:
@Atinux 

No it doesn't. That's why I had a lot of trouble with Nuxt in the beginning!

### The NUXT way to add script, as per the [FAQ](https://nuxtjs.org/faq/):
```
script: [
      { src: 'https://cdnjs.cloudflare.com/ajax/libs/jquery/3.1.1/jquery.min.js' }
    ],
```
vs
### The vue-meta method to add script, as per their [documentation](https://github.com/declandewet/vue-meta#script-object):
```
script: [
      { innerHTML: '{ "@context": "http://schema.org" }', type: 'application/ld+json' }
    ]
```

This is why it was confusing as I was trying to do everything the "vue-meta" way, but only after reading the Nuxt FAQ I realised Nuxt has a different way of doing the script tags.

That's why in this PR I added the link to the FAQ so the users would check the FAQ.

manniL:
@mesqueeb It's the same implementation. You are comparing an inline script vs. a script with `src`attribute here.

PR #652, opened 2018-07-03
Netlify Deployment Guide

aaronransley:
Hey folks,

Here's an initial pass at a guide for deploying to Netlify. I've had a great experience recently and though things are straight forward, I figure an extra guide couldn't hurt in case someone comes looking for it.


Atinux:
Great idea, thanks @aaronransley

PR #662, opened 2018-07-26
Suggestion for showing how to use these routes

MichaelJCole:
I'm starting to learn Vue with Nuxt, and the docs didn't mention how to use the routes generated here.


Atinux:
Thanks @MichaelJCole

PR #690, opened 2018-08-08
Document host retrieval changes

manniL:
see nuxt/nuxt.js#3651


Atinux:
Shouldn't we sort from higher to lower priority?

manniL:
@Atinux We could do this as well but I think as long as the user can extract the priority in a clear manner, it should be fine.

Also (considering the "workflow"), you might want to set up your local connection settings first and override them on the server when in production.

PR #694, opened 2018-08-09
Improve inject documentation

manniL:
related: nuxt/nuxt.js/issues/2233

### DIFF
--- a/en/guide/plugins.md
+++ b/en/guide/plugins.md
@@ -86,22 +86,53 @@ module.exports = {
 
 ## Inject in $root & context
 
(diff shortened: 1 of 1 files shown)


Atinux:
Wow thank you so much @manniL

manniL:
No problem @Atinux :relaxed:

doriandrn:
Thanks for this!
In the latest examples, you have used `new` which is a reserved word in JS and using it as a value might be confusing for some people. I would really change it to `newValue` or something similar.

manniL:
@doriandrn Thanks for letting me know! Will fix it up in a minute :relaxed:

PR #730, opened 2018-08-25
Note about async validate

galvez:


de/api/pages-validate.md
```diff
@@ -16,6 +16,14 @@ validate({ params, query, store }) {
 }
 ```
 
+You can also return promises:
+
+```js
+validate({ params, query, store }) {
+  return new Promise((resolve) => setTimeout(() => resolve()))
```
clarkdo: How about adding a async function？

galvez:
@clarkdo I just realized I edited the `de` file. Fixing.

galvez:
@clarkdo all done.

pi0:
@Atinux Done

qm3ster:
My dudes, shouldn't this be copied to the other "languages" that are currently english? So they stay in sync?

Or even injected into ALL languages?

Atinux:
I think that would be a good idea for translating files easier

qm3ster:
How was your evaluation of GitLocalize? Does it have what it takes to keep the docs properly in sync?
Can we all join there and see the same status, or is it per person?

Atinux:
We will need to find another solution than GitLocalize since the service will be interrupted at the end of this month :/

qm3ster:
How about, all additions must be inserted into all files with a
`<!-- TODO: translate -->`
and all changes must comment out the original and leave a new version to be translated?
Or leave the new version commented, in case we prefer an outdated translation to a regression to english...
So every PR must touch all languages?

qm3ster:
Great, I find out about this great thing only now that it's dead.
And thanks @galvez

mbrookes:
Any more details on the shutdown of gitlocalize? There's no mention on the website, docs site, or blog, and they're still accepting new sign-ups...

PR #733, opened 2018-08-25
Update API/generate, add Russian

qm3ster:


de/api/configuration-generate.md
```diff
@@ -172,28 +162,24 @@ const axios = require('axios')
...
-        return res.data.map((user) => {
-          return {
-            route: '/users/' + user.id,
-            payload: user
-          }
-        })
-      })
-    }
+    routes: () =>
+      axios.get('https://my-api/users').then(res =>
```
Atinux: return statement missing

de/api/configuration-generate.md
```diff
@@ -86,11 +86,7 @@ We add routes for `/users/:id` in `nuxt.config.js`:
 ```js
 module.exports = {
   generate: {
-    routes: [
-      '/users/1',
-      '/users/2',
-      '/users/3'
-    ]
+    routes: ['/users/1', '/users/2', '/users/3']
```
Atinux: I don't think this makes it clearer

de/api/configuration-generate.md
```diff
@@ -172,28 +162,24 @@ const axios = require('axios')
...
 ```
 
 Now we can access the `payload` from `/users/_id.vue` like so:
 
 ```js
-async asyncData ({ params, error, payload }) {
-  if (payload) return { user: payload }
-  else return { user: await backend.fetchUser(params.id) }
-}
+asyncData: async ({ params, error, payload }) =>
```
Atinux: While it might be optimised, I don't think it's a good idea for documentation since it makes it harder to read for a human :)

de/api/configuration-generate.md
qm3ster: What, where?

de/api/configuration-generate.md
qm3ster: It was Prettier that did this, actually. Seemed more than short enough not to split.
I split the french because it uses `/utilisateurs/` instead of `/users/` which imo is a bit much. Maybe the examples should be monolingual, so that if we later have a translation build process they can be imported from one place.

qm3ster:
@Atinux I replied to the comments, tell me what to do please :)

de/api/configuration-generate.md
Atinux: Would like @manniL opinion on this particular comment

qm3ster: @manniL it has come a full circle UwU

manniL: I'd go for something like

```js
async asyncData ({ params, error, payload }) {
  if (payload) {
    return { user: payload }
  }
  const user = await backend.fetchUser(params.id)
  return { user }
}
```

Not all readers know ES6 and short circuits very well, so I'd favor a "more readable" /simple version

qm3ster: Oke, will do.
How do you feel about the original version:
```js
async asyncData ({ params, error, payload }) {
  if (payload) return { user: payload }
  else return { user: await backend.fetchUser(params.id) }
}
```
i'd just change it to arrow notation since `this` isn't in use, and remove the unnecessary `else`
```js
asyncData: async ({ params, error, payload }) => {
  if (payload) return { user: payload }
  return { user: await backend.fetchUser(params.id) }
}
```

manniL: @qm3ster I don't like "inline ifs" actually. Was a fan of them years ago but now I appreciate the extra braces for readability.

This is also aligns with our eslint config.

en/api/configuration-generate.md
```diff
@@ -147,16 +139,14 @@ const axios = require('axios')
 
 module.exports = {
   generate: {
-    routes: function (callback) {
-      axios.get('https://my-api/users')
-      .then((res) => {
-        var routes = res.data.map((user) => {
-          return '/users/' + user.id
+    routes: callback =>
```
qm3ster: Actually @manniL I now see that this may be confusing, because it does call the callback, but also returns a `Promise` which resolves after the callback returns.
Would putting `void axios.get` be too arcane?
Should I just put the body of the `routes` function into a `{}` block so that it returns `undefined`?

de/api/configuration-generate.md
qm3ster: Gotcha, will do like yours. Should the `await` also be on a separate line?
**whispers**: *multiline ifs are a code smell, pull them out into a named function*

en/api/configuration-generate.md
manniL: Not sure if we want to keep a *callback* example for the routes function at all.

qm3ster: That's fine by me. :weary: :ok_hand: 
So I nuke it?
From all languages?

manniL: I'd say so

PR #739, opened 2018-08-29
update port 3000

ZQun:
update port 3000


Atinux:
Actually, we use `nuxt.listen(4000)` for tests :)

PR #770, opened 2018-09-23
Update configuration-css.md

Thivieira:


galvez:
@Thivieira Please pack all future similar modifications into a single PR

PR #783, opened 2018-09-24
Remove section on build.vendor

galvez:
@clarkdo @pi0 @Atinux anything left to update here?


manniL:
I think https://nuxtjs.org/faq/window-document-undefined has `vendor` included as well

PR #785, opened 2018-09-24
Actualize info about css-loader

laSinteZ:
Due to upgrade of CSS loader and my issue https://github.com/nuxt/nuxt.js/issues/3931 I think we should update documentation so we don't mislead newcomers


Atinux:
Thanks @laSinteZ

PR #791, opened 2018-09-26
Updated mistyped nuxt config filename extension

llavre:
Updated mistyped nuxt config filename extension


Atinux:
Thank you @llavre 

It seems that another PR has been merged which do the same update/

PR #806, opened 2018-09-30
Update the JWT External Authenticate pages

liamcharmer:
Making sure that it: 
- The source code reflects the documentation and matches
- The instruction also documents how to logout.


Atinux:
Thanks @liamcharmer

Issue #814, opened 2018-10-02
windows simples quote error to execute node in nodemon -exec


lucasfontesgaspareto:
### Reproduction link

[https://jsfiddle.net/](https://jsfiddle.net/)

### Steps to reproduce

executing the dev script in windows `yarn dev` or `npm run dev`

### What is expected ?

```
C:\Users\***\docs>npm run dev

> nuxt-docs@1.0.0 dev C:\Users\***\docs
> nodemon --watch *.js --exec "node api.js"

[nodemon] 1.18.4
[nodemon] to restart at any time, enter `rs`
[nodemon] watching: *.js
[nodemon] starting `node api.js`
Building files...
Building menu...
Building languages...
Fetching releases...
Watch files changes...
Server listening on localhost:4000
```

### What is actually happening?

```
C:\Users\***\docs>npm run dev

> nuxt-docs@1.0.0 dev C:\Users\***\docs
> nodemon --watch *.js --exec 'node api.js'

[nodemon] 1.18.4
[nodemon] to restart at any time, enter `rs`
[nodemon] watching: *.js
[nodemon] starting `'node api.js'`
''node' não é reconhecido como um comando interno
ou externo, um programa operável ou um arquivo em lotes.
[nodemon] app crashed - waiting for file changes before starting...
```

<div align="right"><sub><em>This bug report is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/docs/issues/c113">#c113</a>)</em></sub></div>

PR #816, opened 2018-10-02
fix: #814

lucasfontesgaspareto:


galvez:
@lucasfontesgaspareto I see that you changed the lang.json format to YAML -- will that still work?

Would be better to keep consistency: https://github.com/nuxt/docs/blob/master/en/lang.json

cc @Atinux @alexchopin

galvez:
@manniL do you think we should wait until your refactor branch is done?

manniL:
@galvez I think it's fine as only the package.json will possibly have a conflict.

PR #830, opened 2018-10-07
Update installation.md

negope:


manniL:
Hey @negope 👋 
This is no typo but intended. Read more about npx [here](https://alligator.io/workflow/npx/)

PR #853, opened 2018-10-11
Doc Update: Added hint as to why arrow functions don't have there own 'this'

### LINKED ISSUE #849: vuex: this is undefined in arrow function
### What problem does this feature solve?

A short hint on https://nuxtjs.org/guide/vuex-store/ that `this` isn't available if you use arrow functions in file-split store modules would be nice.

Related: https://github.com/nuxt/nuxt.js/issues/4047

<div align="right"><sub><em>This feature request is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/docs/issues/c120">#c120</a>)</em></sub></div>

Esslamb:
Tried to keep the hint as short as possible let me know if you want me to remove a few lines! I placed is under Module Files sections as I felt it would be at this point that most people would run into issues with not having access to ```this``` because of using arrow functions.

Resolves #849

### DIFF
--- a/en/guide/vuex-store.md
+++ b/en/guide/vuex-store.md
@@ -195,6 +195,8 @@ export default {
 
 You can optionally break down a module file into separate files: `state.js`, `actions.js`, `mutations.js` and `getters.js`. If you maintain an `index.js` file with state, getters and mutations while having a single separate file for actions, that will also still be properly recognized.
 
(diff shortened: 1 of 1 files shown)


manniL:
Thanks for your contribution! 🙏 ☺️

PR #875, opened 2018-10-18
feat: add advanced nginx example

pimlie:
Added an advanced nginx example for when you use Nuxt to generate all (dynamic) routes of your project so nginx can serve mostly static files. But I also added a caching proxy to a Nuxt server so if an url is not (yet) generated the Nuxt server will generate the page for that url and nginx will cache it until its invalidated or purged.

The example doesnt explain how everything works, its more to show what is possible and if you want to know exactly how and why then you should RTFM (of nginx mostly)

### DIFF
--- a/en/faq/nginx-proxy.md
+++ b/en/faq/nginx-proxy.md
@@ -39,7 +39,7 @@ server {
 
 If you have a high volume website with regularly changing content, you might want to benefit from Nuxt generate capabilities and [nginx caching](https://www.nginx.com/blog/nginx-caching-guide).
 
-Below is an exampe configuration. Keep in mind that:
+Below is an example configuration. Keep in mind that:
 - root folder should be the same as set by [configuration generate.dir](/api/configuration-generate#dir)
 - expire headers set by Nuxt are stripped (due to the cache)
 - both Nuxt as nginx can set additional headers, it's advised to choose one (if in doubt, choose nginx)
--- a/en/faq/nginx-proxy.md
+++ b/en/faq/nginx-proxy.md
@@ -39,10 +39,20 @@ server {
 
 If you have a high volume website with regularly changing content, you might want to benefit from Nuxt generate capabilities and [nginx caching](https://www.nginx.com/blog/nginx-caching-guide).
 
-Below is an exampe configuration which you will most likely need to adapt to your needs.
+Below is an exampe configuration. Keep in mind that:
+- root folder should be the same as set by [configuration generate.dir](/api/configuration-generate#dir)
+- expire headers set by Nuxt are stripped (due to the cache)
+- both Nuxt as nginx can set additional headers, it's advised to choose one (if in doubt, choose nginx)
+- if your site is mostly static, increase the `proxy_cache_path inactive` and `proxy_cache_valid` numbers
+
+If you don't generate your routes but still wish to benefit from nginx cache:
+- remove the `root` entry
+- change `location @proxy {` to `location / {`
+- remove the other 2 `location` entries
+
 
 ```nginx
-proxy_cache_path  /PATH/TO/YOUR/CACHE levels=1:2 keys_zone=nuxt-cache:25m max_size=1g inactive=60m use_temp_path=off;
+proxy_cache_path  /data/nginx/cache levels=1:2 keys_zone=nuxt-cache:25m max_size=1g inactive=60m use_temp_path=off;
 
 map $sent_http_content_type $expires {
     "text/html"                 1h; # set this to your needs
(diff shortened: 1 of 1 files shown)


en/faq/nginx-proxy.md
```diff
@@ -35,6 +35,80 @@ server {
...
+    listen          80;             # the port nginx is listening on
+    server_name     your-domain;    # setup your domain here
+
+    gzip            on;
+    gzip_types      text/plain application/xml text/css application/javascript;
+    gzip_min_length 1000;
+
+    charset utf-8;
+
+    root /var/www/YOUR_NUXT_DIST_FOLDER
```
pi0: I suggest making a more clear example that this is the root `dist` folder or the on onside `.nuxt`

en/faq/nginx-proxy.md
```diff
@@ -35,6 +35,80 @@ server {
...
+    "text/html; charset=utf-8"  1h; # set this to your needs
+    default                     7d; # set this to your needs
+}
+
+server {
+    listen          80;             # the port nginx is listening on
+    server_name     your-domain;    # setup your domain here
+
+    gzip            on;
+    gzip_types      text/plain application/xml text/css application/javascript;
```
pi0: What about static assets like images?

en/faq/nginx-proxy.md
```diff
@@ -35,6 +35,80 @@ server {
...
+# Using nginx with generated pages and a caching proxy as fallback
+
+If you have a high volume website with regularly changing content, you might want to benefit from Nuxt generate capabilities and [nginx caching](https://www.nginx.com/blog/nginx-caching-guide).
+
+Below is an exampe configuration which you will most likely need to adapt to your needs.
+
+```nginx
+proxy_cache_path  /PATH/TO/YOUR/CACHE levels=1:2 keys_zone=nuxt-cache:25m max_size=1g inactive=60m use_temp_path=off;
+
+map $sent_http_content_type $expires {
```
pi0: Is this for serving generated version only?

en/faq/nginx-proxy.md
pimlie: That part I copied from the default nginx proxy example :)

en/faq/nginx-proxy.md
```diff
@@ -35,6 +35,80 @@ server {
...
 ```
 
+# Using nginx with generated pages and a caching proxy as fallback
+
+If you have a high volume website with regularly changing content, you might want to benefit from Nuxt generate capabilities and [nginx caching](https://www.nginx.com/blog/nginx-caching-guide).
+
+Below is an exampe configuration which you will most likely need to adapt to your needs.
+
+```nginx
+proxy_cache_path  /PATH/TO/YOUR/CACHE levels=1:2 keys_zone=nuxt-cache:25m max_size=1g inactive=60m use_temp_path=off;
```
pi0: Trust me. Most people love to copy paste it :D Add a default real path.

en/faq/nginx-proxy.md
pimlie: No, you could use the proxy_cache as well if you dont use generated pages. Actually the only part for the generated pages is setting the root and adding `try_files $uri/index.html` ;)

en/faq/nginx-proxy.md
pimlie: Ok, will use the same one as the nginx docs

en/faq/nginx-proxy.md
pimlie: Or shall I write [nuxt.options.generate.dir] ?

en/faq/nginx-proxy.md
pimlie: But actually its not really useful to gzip images, until you are using tiff's or bmp's images are compressed already so it would be a waste to spent cpu cycles to try to gzip them.

en/faq/nginx-proxy.md
pimlie: I will change the description of the example to point that out, thanks

en/faq/nginx-proxy.md
pi0: Still confusing for copy-paste users IMO :))

en/faq/nginx-proxy.md
pi0: Yes! But doesn't it override expires cache headers sent from Nuxt? (Those that sent from nuxt.config). I mean it's actually better idea to delegate such configurations to nginx layer. But users should be aware of that.

en/faq/nginx-proxy.md
pi0: You are completely right @pimlie :D

pimlie:
Hope this is better!

manniL:
Thank you for you contribution :+1:

Issue #879, opened 2018-10-18
Fix source links


manniL:
After the monorepo refactor source links are off. We should point them again to the right spots. Example: [internals](https://nuxtjs.org/api/internals#nuxt-internals)

<div align="right"><sub><em>This bug report is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/docs/issues/c133">#c133</a>)</em></sub></div>

stuin:
I can get this. It looks like all of them are in the internals section? Should be pretty quick for that.

PR #880, opened 2018-10-18
Updated all source code urls

stuin:
Issue #879 
Made use of some "Find and Replace"
Multiple rounds, one for each package.


manniL:
@stuin Thanks a bunch! :relaxed:

PR #885, opened 2018-10-19
fix layoutTransition property

mxrlkn:


manniL:

PR #888, opened 2018-10-20
feat: huge doc revamp

manniL:


en/api/index.md
```diff
@@ -24,4 +26,4 @@ export default {
 }
 ```
 
-<div class="Alert Alert--orange">You do **NOT** have access of the component instance through `this` inside `asyncData` because it is called **before initiating** the component.</div>
+<p class="Alert Alert--orange"><b>Warning:</b> You **don't** have access of the component instance through `this` inside `asyncData` because it is called **before initiating** the component.</p>
```
galvez: @manniL is there different rendering for `<b>` and `**`?

manniL: I don't think so! (example: https://nuxtjs.org/faq/github-pages#how-to-deploy-on-github-pages-) 

Good point.
Still need to find out why markdown in HTML tags isn't properly resolved in my local setup but on the live page (cc @Atinux :pray:)

en/guide/configuration.md
```diff
@@ -43,13 +43,13 @@ This option lets you define all default meta tags for your application.
...
 
-This option lets you customize the loading component Nuxt.js loads by default.
+This option lets you customize the loading component Nuxt.js uses by default.
 
 [Documentation about `loading` integration](/api/configuration-loading)
 
 ### modules
 
-This option lets you add Nuxt modules to your project.
+Wit this option you can add Nuxt modules to your project.
```
galvez: @manniL typo here.

manniL: Thanks!

en/api/index.md
Atinux: Good question, I will switch to another Markdown to HTML parser soon, you can keep HTML like you did.

PR #893, opened 2018-10-22
Fix: link within HTML tag

NeverBehave:


NeverBehave:
I have noticed that an extra `<p>` tag will be added between the `<div>`, which causes the the text color the same as the background color

image

What I suggest is that maybe define the `text-color` at `Alert` class for tag `p` at file https://github.com/nuxt/nuxtjs.org/blob/master/assets/scss/theme/_content.scss in line 200.

manniL:
Hey! Thanks for the contribution. As this affects all Alerts, I've created a PR to fix them all and changed the CSS properly as well.

Closed in favor of #896

PR #896, opened 2018-10-23
Update default babel config description for nuxt@2.x

williamchong007:


manniL:
Good point! 

PR #898, opened 2018-10-23
Update configuration-css.md

Lukang689:
Correct grammatical errors.


manniL:
Hey @Lukang689! Please stick with `export default` as the `nuxt.config.js` supports ESM syntax since Nuxt 2

PR #902, opened 2018-10-23
Add warning note about isClient and isServer

XanderLuciano:
Add short warning note about `isClient`/`isServer` in build extend to deter developers from thinking it is deprecated and should be replaced with `process.client`/`process.server`, which would return undefined and therefore cannot be used.

There appearrs to be a large amount of misinformation about this property, leading to a lot of incorrect configurations that *appear* to be correct because replacing `isClient` with `process.client` will disable linting, making it appear as if there are no errors, when the reality is they have just disabled linting.

### DIFF
--- a/en/api/configuration-build.md
+++ b/en/api/configuration-build.md
@@ -88,6 +88,8 @@ The extend is called twice, one time for the server bundle, and one time for the
 1. webpack config object,
 2. object with the following keys (all boolean): `isDev`, `isClient`, `isServer`, `loaders`.
 
+*note: The `isClient` and `isServer` keys in extend are separate from the keys available in context. `process.client` and `process.server` are undefined at this point and cannot be used.*
+
 Example (`nuxt.config.js`):
 
 ```js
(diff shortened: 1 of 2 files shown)


XanderLuciano:
Additionally we may want to consider adding a note on the context API page as well.  
https://nuxtjs.org/api/context#context

XanderLuciano:
Great idea! I didn't realize I wasn't strictly limited to markdown 🙂 @manniL

manniL:
Could you use the `orange` version and use `Warning:` as bold text? 

It's nowhere written but I try to establish a convention there 🙈 
Sorry 😅

XanderLuciano:
@manniL Thanks for the example! Maybe some styling guides could be added to the repo's wiki just so there's a place to reference them easily (and copy paste from). 😉

manniL:
@XanderLuciano That's on the list ☺️ Thanks for your feedback! 

Please make sure the HTML (and MD in HTML) is rendered properly by including another line break (as in the example I've linked). Depending on the MD spec there are different constraints for parsing the content correctly but two linebreaks should be always safe.

XanderLuciano:
@manniL 2 blank lines between obejcts that is? (for a total of 3 line breaks?)

Hopefully I understood that correctly.

manniL:
@XanderLuciano Almost right. I fixed it :relaxed:

XanderLuciano:
Ah, that makes sense now 🙂 thanks!

manniL:
Thank you! :relaxed: 
Can you add an alert to the context page as well? :relaxed: 
(Maybe sth. like "this is not the context passed in to build.extend")

PR #908, opened 2018-10-25
Explaining function behaviour in a clearer way

davidemorotti:
Hi, I was struggling at understanding why the asyncData function doesn't always load server-side.
Basing on this explanation: https://github.com/nuxt/nuxtjs.org/issues/106#issuecomment-360808831 (who made me understand how nuxtjs works), I decided to try a different way of describing the function, hopefully helping peoples with the same problem.


manniL:

PR #910, opened 2018-10-25
Update vuex-store.md

davidemorotti:
Clearer explanation, sometimes user (like me) just interpret the phrases "before rendering" same as "server-side".


manniL:
Mentioning this in the fetch API page seems sufficient for me.

PR #918, opened 2018-10-26
Add comma at the end of element in postcss array

runyasak:
I've seen that there have no comma at the end of elements in postcss array.
So, I add in docs on Faq page.


manniL:

PR #929, opened 2018-10-29
Update middleware doc

WangHansen:
Explain that middleware only get executed once and from server when page reload


manniL:
Yeah, you are right, we should make clear "when SSR happens" somewhere. But maybe there is a better place for that :thinking:

PR #937, opened 2018-10-30
Fix plugins/i18n.js section

alexbonhomme:
Replaces `store` by `app` in context params and removes useless `inject` argument


manniL:

PR #943, opened 2018-11-01
en: Fix url of en/api/internals.md

aytdm:
@manniL 
Hi,
Some links are 404, so I fixed that.
Could you review this PR? correct URLs?

Fix url and link text in brackets below:
+ renderer.js
+ options.js


manniL:

While converting to a monorepo this happens fast 🙈 
Pretty sure that won't be the last time URLs will change

aytdm:
Thank you for your review :relaxed:

PR #944, opened 2018-11-01
en: Fix url

aytdm:
related: #943


manniL:
@aytdm Thank you! While you are on it, could you go through the other languages as well? That'd be lovely! 🙏

aytdm:
@manniL Sure! I think that I fixed all other languages. Please check :pray:

manniL:
Many thanks! 👍

PR #946, opened 2018-11-02
vue.config change productTip for productionTip

jasonujmaalvis:
Update `productTip` to `productionTip`. 

I think it's a typo as `productTip` doesn't look like it's a vue.config option.


manniL:
Many thanks! 👍

PR #955, opened 2018-11-06
Update views.md

napestershine:


manniL:
@napestershine Thanks for the PR. Already corrected the URL in #956 though. It should link to the *original code* as well, not to the example 👍 

Thanks anyway!

PR #971, opened 2018-11-11
fix: github links

soata:

PR #972, opened 2018-11-11
fix invalid links

soata:
repull-request #971 

change 'https://github.com/nuxt/nuxt.js/tree/dev/packages/app/template' to 'https://github.com/nuxt/nuxt.js/tree/dev/packages/vue-app/template'


manniL:
:+1: Thanks!

PR #979, opened 2018-11-13
nuxt/docs does not use npm but Yarn

inouetakuya:
nuxt/docs does not use npm but Yarn.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,8 +11,8 @@
 Start a dev server on `localhost:4000`:
 
 ```bash
-npm install
-npm run dev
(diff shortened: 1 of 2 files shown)


inouetakuya:
https://github.com/nuxt/docs/commit/94761a8c8efd2f99ebf101112850ffd6030fbea9#diff-32607347f8126e6534ebc7ebaec4853d

package-lock.json had been added.

---

> nuxt/docs does not use npm but Yarn.

I have misunderstood. I will close this PR.

PR #980, opened 2018-11-13
ja: Install textlint for Japanese

inouetakuya:
Japanese translation try to use [textlint](https://github.com/textlint/textlint).


aytdm:
いいですね！ありがとうございます😆

inouetakuya:
@aytdm CC: @Atinux @potato4d 

Thanks. This is try. Let's improve translation process!! 💪

PR #991, opened 2018-11-17
feat(api, en): devtools option

mottox2:
Related PR: nuxt/nuxt.js#4357


manniL:
@mottox2 Not "100% correct" because not using the flag doesn't disable the dev-tools if already activated through `nuxt.config.js` or otherwise

mottox2:
@manniL Thanks for review. I fiexed it.

manniL:
Thank you ☺️

PR #993, opened 2018-11-19
Update asyncData docs

podlebar:
Updated docs as the callback-based asyncData, fetch or middleware calls are deprecated since 2.3.1.


manniL:
Thanks. missed that somehow! :+1:

PR #1017, opened 2018-12-01
Replace process.BROWSER_BUILD and process.browser with process.client

korigod:
`process.BROWSER_BUILD` was removed in release v1.0, so now the docs are incorrect.


manniL:
Please replace it with `process.client` instead :relaxed:
It's preferred over `process.browser`

korigod:
@manniL, thanks, I've replaced it everywhere.

PR #1019, opened 2018-12-03
ZEIT Now deployment configuration update

Igloczek:
Current docs are describing how to setup deployment for Now V1, while it's already a legacy version and it doesn't work anymore on Now V2.

For now, I omitted the part of setting up the deployment of a universal app with SSR, because still need to figure out how to do it.


Igloczek:
@manniL updated :)

PR #1034, opened 2018-12-11
updated build:templates args

hecktarzuli:
The obj that's passed to build:templates hook actually has the prop templatesFiles not templateFiles


manniL:

PR #1036, opened 2018-12-12
Bugs found in YAML Front-Matter and JSON menu.

renoirb:
Here are two snags that broke running in development.

Maybe we could setup prettier to check all YAML/JSON.


manniL:
Thanks! ☺️

PR #1039, opened 2018-12-13
Replace `nuxt.showOpen` with `nuxt.showReady`

mubaidr:
`nuxt.showOpen` is not a function

### DIFF
--- a/de/api/configuration-dev.md
+++ b/de/api/configuration-dev.md
@@ -43,8 +43,7 @@ app.use(nuxt.render)
 if (config.dev) {
   new Builder(nuxt).build()
   .catch((error) => {
-    console.error(error)
-    process.exit(1)
+    throw error
   })
 }
 
--- a/de/api/configuration-dev.md
+++ b/de/api/configuration-dev.md
@@ -42,9 +42,6 @@ app.use(nuxt.render)
 // Build only in dev mode
 if (config.dev) {
   new Builder(nuxt).build()
-  .catch((error) => {
-    throw error
-  })
 }
 
 // Listen the server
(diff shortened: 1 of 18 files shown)


mubaidr:
Removed `nuxt.showReady` and replace process.exit with throw.

mubaidr:
You are right, we don't need to catch error at all. This: 
```bash
(node:20732) [DEP0018] DeprecationWarning: Unhandled promise rejections are depr
ecated. In the future, promise rejections that are not handled will terminate th
e Node.js process with a non-zero exit code.
```

after all we are just logging and exiting process.

ja/api/configuration-dev.md
```diff
@@ -42,15 +42,11 @@ app.use(nuxt.render)
...
-  .catch((error) => {
-    console.error(error)
-    process.exit(1)
-  })
 }
 
 // サーバーを Listen する
 app.listen(port, '0.0.0.0').then(() => {
-  nuxt.showOpen()
+  console.log(`Server listning on port: ${port}`)
```
manniL: Typo here
```suggestion
  console.log(`Server listening on port: ${port}`)
```

mubaidr: 🤦‍♂️

de/api/configuration-dev.md
```diff
@@ -23,31 +23,31 @@ Example:
...
+};
 ```
 
 `server.js`
 
 ```js
-const { Nuxt, Builder } = require('nuxt')
-const app = require('express')()
-const port = process.env.PORT || 3000
+const { Nuxt, Builder } = require('nuxt');
```
manniL: We are coming closer! However, please use semicolon-less style and only include necessary changes :relaxed:

mubaidr:
@manniL please https://github.com/nuxt/docs/pull/1065

PR #1040, opened 2018-12-13
build.transpile usage clarification

leroyshirtoFH:
Add additional clarification detailing the use of build.transpile.
It was not clear that the string could be the name of the package to transpile.


manniL:

PR #1046, opened 2018-12-18
Create eventsource-polyfill.md

Lukang689:
When project runs in some browser, webpack-hot-middleware's client requires EventSource to work. We should include a polyfill if we want to support this browser.

### DIFF
--- a/en/faq/eventsource-polyfill.md
+++ b/en/faq/eventsource-polyfill.md
@@ -1,0 +1,29 @@
+---
+title: A polyfill to support browser
+description: How to include a polyfill to support browser?
+---
+
+# How to include a polyfill to support browser?
+
+```
+You are running Vue in development mode.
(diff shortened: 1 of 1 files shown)


en/faq/eventsource-polyfill.md
```diff
@@ -0,0 +1,29 @@
...
+```
+In `nuxt.config.js`：
+
+```
+ build: {
+    /*
+    ** You can extend webpack config here
+    */
+    extend(config, ctx) {
+      if(ctx.idDev) {
```
clarkdo: Should be `isDev`

Lukang689: Oh,my carelessness.Already modified.

Atinux:
Hi @Lukang689 

We won't have to write this with this PR https://github.com/nuxt/nuxt.js/pull/4600

Thanks for the idea :)

baturali:
Hi, when i run in development mode my Nuxt project this error comes on Edge and IE browsers. Any help? 
webpack-hot-middleware's client requires EventSource to work. You should include a polyfill if you want to support this browser: https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events#Tools

PR #1047, opened 2018-12-18
fix nuxt command `bin/nuxt` to `yarn nuxt`

suusan2go:
`bin/nuxt` dose not exist.


manniL:
@suusan2go Are there other places where this is used as well? 🤔

manniL:
Thank you :relaxed:

PR #1051, opened 2018-12-19
en: fix incorrect pluralization

scott-joe:


manniL:
Many thanks :relaxed:

PR #1054, opened 2018-12-20
en: add routeNameSplitter

ricardogobbosouza:


en/api/configuration-router.md
```diff
@@ -31,6 +31,22 @@ When `base` is set, Nuxt.js will also add in the document header `<base href="{{
 
 > This option is given directly to the vue-router [base](https://router.vuejs.org/api/#base).
 
+## routeNameSplitter
+
+- Type: `String`
+- Default: `'-'`
+
+You may want to change route name splitter created by Nuxt.js. You can do so via the `routeNameSplitter` option.
```
manniL: Could you provide more detail what exactly the `route name splitter` is/what changing it causes?

ricardogobbosouza: What do you think now?

manniL:
Thank you ☺️

PR #1069, opened 2018-12-28
add note about scrollToTop: false

galvez:
Re https://github.com/nuxt/nuxt.js/pull/4564


manniL:
Should be in branch 2.4 instead of master.

PR #1091, opened 2019-01-10
Fix globalName property type error

schaoss:
Modify property `globalName`'s type from `Object` to `String`


manniL:

Issue #932, opened 2018-10-29
Summary of context usage


manniL:
### What problem does this feature solve?

Page: https://nuxtjs.org/api/context#context

Summarize usage, what the context is for and where the context is available 

<div align="right"><sub><em>This feature request is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/docs/issues/c140">#c140</a>)</em></sub></div>

PR #1092, opened 2019-01-10
Include information for context

Relequestual:
Copied hidden information in the same file to be presented on the site docs.
Following the same style as https://github.com/nuxt/docs/blob/master/en/api/index.md

Looks to partially fix https://github.com/nuxt/docs/issues/932

---
I was trying to find where `this.$route` for example is defined in the documentation. I couldn't find it, and I know this doesn't fix that issue, hence suggesting this is a partial fix. I don't feel I'd know how to best explain this.


Atinux:
Thanks @Relequestual

PR #1096, opened 2019-01-11
List indentation fixes

buddyy93:
i'm sorry i used prettier formatter on my vs code and when i saved theh file the extension formatted the whole file and i didn't notice the indentation changes on the list, this update should fix it, thank you


manniL:
Thanks! Could you apply that to other files (in the other languages) as well?

buddyy93:
> Thanks! Could you apply that to other files (in the other languages) as well?

hi, i'd love to but i don't see any indentation problems on other files (in other langauges) because i edited the indonesian installation guide only

PR #1098, opened 2019-01-12
en: modern mode with nuxt generate

mornir:
After watching Sebastien Chopin (https://youtu.be/-Z1hI4Wxblo), I thought it would be nice to have that information in the docs.


manniL:
Thanks :relaxed:

PR #1109, opened 2019-01-17
Update vuex-store.md

Cjumelin:
Dead link, change it just for fr lang set


manniL:
@Cjumelin Thanks ☺️ Could you also change the other vuex links on that page?

PR #1125, opened 2019-01-24
Update modules.md

felixdenoix:
Conjugaison is hard :) ( amazing work with Nuxt btw 👍)


manniL:
Thank you :raised_hands:

PR #1153, opened 2019-02-03
Add 0.0.0.0 usage clarification

Eido95:
Host `0.0.0.0` have different meanings between different contexts.

I added a clarification to what host `0.0.0.0` is designated to in this specific context.


manniL:
Many thanks :+1:

PR #1154, opened 2019-02-04
doc: functional babel presets

clarkdo:


manniL:
@clarkdo Updated the order of info (because I guess most ppl want to change the targets and only a few want to use completely custom presets ☺️ )

clarkdo:
It looks perfect！

PR #1164, opened 2019-02-14
Fixed #scrollbehavior anchor link.

unr:
Previously linked to #scrollBehavior which was invalid.

### DIFF
--- a/en/api/pages-scrolltotop.md
+++ b/en/api/pages-scrolltotop.md
@@ -25,4 +25,4 @@ export default {
 
 Conversely, you can manually set `scrollToTop` to `false` on parent routes as well.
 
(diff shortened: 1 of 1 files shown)


unr:
No problem @farzadso :)

manniL:
@unr Thanks :relaxed: 
Is that problem present in other languages as well?

farzadso:
@manniL Just checked. Yes, It exists in other languages too. I think a global replacement is needed.

unr:
Could alternatively update the page to use #scrollBehavior instead 🤔

manniL:
As you wish :relaxed:

unr:
Hmmm, I was hoping to just make that #scrollBehavior change to make it a simpler fix.

Seems that the anchor tag is being generated, not something I can one-character-fix. :)

PR #1165, opened 2019-02-15
Update postcss-plugins.md

core01:
If there will be postcss.config.js, nuxt.config.js postcss property will be ignored

Reproduce link: https://github.com/core01/postcss-simple-vars-nuxt-reproduction

### DIFF
--- a/en/faq/postcss-plugins.md
+++ b/en/faq/postcss-plugins.md
@@ -5,7 +5,7 @@ description: How to add PostCSS plugins?
 
 # How to add PostCSS plugins?
 
-### Default Method
+### Recommended Method
 Delete the `postcss.config.js` in your project directory (if existent). Then, in your `nuxt.config.js` file add the following:
 
 ```js
--- a/en/faq/postcss-plugins.md
+++ b/en/faq/postcss-plugins.md
@@ -6,18 +6,6 @@ description: How to add PostCSS plugins?
 # How to add PostCSS plugins?
 
 ### Default Method
-Use `postcss.config.js`, for example:
-
-```
-const join = require('path').join
-const tailwindJS = join(__dirname, 'tailwind.js')
-
-module.exports = {
-  plugins: [require('tailwindcss')(tailwindJS), require('autoprefixer')]
-}
-```
-
-### Modern Method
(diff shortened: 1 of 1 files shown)


core01:
@farzadso done, check it please!

farzadso:
Thanks @core01 

It looks much better now.

I would also suggest adding back-ticks before and after `postcss.config.js`

core01:
@farzadso back-ticks added

en/faq/postcss-plugins.md
```diff
@@ -5,7 +5,20 @@ description: How to add PostCSS plugins?
 
 # How to add PostCSS plugins?
 
-In your `nuxt.config.js` file:
+### Default Method
```
manniL: I'd probably switch the two heading. The "default" should be using `nuxt.config.js`. The "legacy" or "traditional" way is a postcss.config.js

core01: @manniL will it good if I also swap two methods? Then the first will be default (nuxt.config.js) and the second (postcss.config.js) ?

manniL: Yes, please. :relaxed:

en/faq/postcss-plugins.md
```diff
@@ -5,7 +5,8 @@ description: How to add PostCSS plugins?
...
```
manniL: Maybe
```suggestion
### Recommended Method
```

en/faq/postcss-plugins.md
```diff
@@ -5,7 +5,8 @@ description: How to add PostCSS plugins?
...
+Delete the `postcss.config.js` in your project directory (if existent). Then, in your `nuxt.config.js` file add the following:
```
manniL:
```suggestion
If present, rename or delete the `postcss.config.js` in your project directory. Then, in your `nuxt.config.js` file add the following:
```

core01: @farzadso what do you think about it?

farzadso: @core01 Couldn't agree any more. It's got much better sentencing.

manniL:
Thanks ☺️

PR #1169, opened 2019-02-23
docs: improve sentence grammar

PedroPovedaQ:
After the altered sentence we have "It will build a Static Generated Vue.js Application." Which has a better grammatical style. Now the preceding sentence follows this standard.


manniL:
Thank you :relaxed:

PR #1174, opened 2019-03-02
ja: Add files not exists under ja directory and Update ja/**/menu.json

aytdm:
@inouetakuya @potato4d I added  these files not exists under ja directory for translation.


aytdm:
@inouetakuya Oh, I missed it. Thank you for pointing that! I fixed and PTAL:bow:

PR #1185, opened 2019-03-05
Update plugins.md

errietta:
Without doing `Vue.use` inside an exported function in a plugin, any injections will not happen. Update the documentation to point this out.


Atinux:
Actually it will work @errietta 

Do you have a reproduction where it does not work?

farzadso:
I would also like to see an example as I'm simply using `Vue.use()` ins all my plugins without any hassle.

PR #1186, opened 2019-03-05
added extended CSP example

NextStepGuru:
I've added a complete CSP example with support for Google Analytics, Sentry.io, and LogRocket.io support.


Atinux:
Thanks @NextStepGuru

PR #1193, opened 2019-03-07
en: nuxtignore

clarkdo:


de/api/configuration-ignore.md
```diff
@@ -12,10 +41,13 @@ description: Define the ignore files for your Nuxt.js application
 
 By default all files which start with `-` will be ignored, such as `store/-foo.js` and `pages/-bar.vue`. This allows for co-locating tests, utilities, and components with their callers without themselves being converted into routes, stores, etc.
 
+**Note:** This option will be deprecated in Nuxt.js 3, we recommend using `.nuxtignore` file instead
```
manniL:
```suggestion
**Note:** This option will be deprecated in Nuxt.js 3. We recommend using a `.nuxtignore` file instead.
```

de/api/configuration-ignore.md
```diff
@@ -12,10 +41,13 @@ description: Define the ignore files for your Nuxt.js application
...
+
 # The ignore Property
 
 - Type: `Array`
 - Default: `['**/*.test.*']`
 
 > More customizable than `ignorePrefix`: all files matching glob patterns specified inside `ignore` will be ignored in building.
 
+**Note:** This option will be deprecated in Nuxt.js 3, we recommend using `.nuxtignore` file instead
```
manniL:
```suggestion
**Note:** This option will be deprecated in Nuxt.js 3. We recommend using a `.nuxtignore` file instead.
```

de/api/configuration-ignore.md
```diff
@@ -3,6 +3,35 @@ title: "API: The ignore Property"
 description: Define the ignore files for your Nuxt.js application
 ---
 
+# .nuxtignore
+
+You can use `.nuxtignore` to let Nuxt.js building phase ignore `layout`, `page`, `store` and `middleware` files in your project’s root directory (`rootDir`). The `.nuxtignore` file is file wih same specification like `.gitignore` and `.eslintignore` in which each line is a glob pattern indicating which files should be ignores.
```
manniL:
```suggestion
You can use a `.nuxtignore` file to let Nuxt.js ignore `layout`, `page`, `store` and `middleware` files in your project’s root directory (`rootDir`) during the build phase.
The `.nuxtignore` file is subject to the same specification as `.gitignore` and `.eslintignore` files, in which each line is a glob pattern indicating which files should be ignored.
```

manniL: Some info that'd be good to have in there:

* Is the `.nuxtignore` file only interpreted/found when it's in the project root directory (as .gitignore files can be nested in subfolders)
* Are there CodeSandbox examples?

clarkdo: Yes. Only works for root dir.
Example needs to be added

de/api/configuration-ignore.md
```diff
@@ -3,6 +3,36 @@ title: "API: The ignore Property"
 description: Define the ignore files for your Nuxt.js application
 ---
 
+# .nuxtignore
```
manniL: Why adding it to de as well? :thinking:

clarkdo: The original pr is opened by adding de incorrectly, I pushed en just now.

Since this page in de is English, do I need to remove it ?

manniL: ah!

no, you can keep it :relaxed:

PR #1256, opened 2019-03-18
Add FAQ about loading audio files with Webpack

mornir:
I hard a hard time finding how to achieve this in Nuxt 2.
Please correct me if this is not the optimal way 🙇‍♂️

I tried setting `transformAssetUrls` in the extend function. It did update the config, but I still couldn't import audio file with the `src` attribute.

```js
export default {
  build: {
    extend (config, { isClient, loaders: { vue } }) {

      // This doesn't work for me.
      vue.transformAssetUrls.audio= 'src'

      config.module.rules.push({
        test: /\.(ogg|mp3|wav|mpe?g)$/i,
        loader: 'file-loader',
        options: {
          name: '[path][name].[ext]',
        },
      })
  
    }
  }
}
```

### DIFF
--- a/en/faq/webpack-audio-files
+++ b/en/faq/webpack-audio-files
@@ -5,7 +5,7 @@ description: How to extend Webpack config to load audio files
 
 #  How to extend Webpack config to load audio files
 
-Audio files should be processed by `file-loader`. This loader is already includes in the default Webpack config, but it is not set to handle audio files. You need to extend its default config in `nuxt.config.js`:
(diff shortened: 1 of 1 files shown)


mornir:
Yes, it would be nice if the processing of audio files was already set up in the base Webpack config of Nuxt. (Even if loading audio files from the assets folder is much less common than for images.)

Atinux:
Thanks @mornir 

I believe a module like @nuxtjs/audio could be even better honestly :)

PR #1261, opened 2019-03-21
Fix typo - add assignment operator

danielroe:


kevinmarrec:
Thanks @danielroe !

@manniL This PR needs to target `2.5` right ? I don't now why we can see my commit inside this PR.

danielroe:
Yes - my apologies! Let's see if I can untangle.

PR #1265, opened 2019-03-22
Grammar fix in last paragraph

BenjaminWalsh:
I've fixed a grammatical error in the page so it reads slightly better.


manniL:
Thanks!

PR #1274, opened 2019-03-26
fix: describe about render.compressor

1000ch:
- remove `(or a falsy value)` because [falsy value will disable compressor option](https://github.com/nuxt/nuxt.js/blob/ef41e205e6787b784ab5907d99652b9d74891c6c/packages/server/src/server.js#L78-L86)
- tweak Japanese translation


manniL:
Thanks! 👍

PR #1283, opened 2019-03-31
Add nuxt.ready (Nuxt 2.5 new API) to docs sample source codes.

potato4d:
Added nuxt.ready to the code that can not be assumed to work as it has been.

### DIFF
--- a/en/api/configuration-dev.md
+++ b/en/api/configuration-dev.md
@@ -37,16 +37,21 @@ const port = process.env.PORT || 3000
 // We instantiate Nuxt.js with the options
 let config = require('./nuxt.config.js')
 const nuxt = new Nuxt(config)
-app.use(nuxt.render)
 
-// Build only in dev mode
-if (config.dev) {
-  new Builder(nuxt).build()
-}
+// Await for initialize Nuxt
+nuxt.ready()
+.then(() => {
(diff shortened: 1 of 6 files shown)


manniL:
This change will be reverted in https://github.com/nuxt/nuxt.js/pull/5413 :relaxed: 	

Not sure if we should add it anyway? (cc @pi0)

Atinux:
Since it has been reverted, I am closing this PR.

Thank you very much @potato4d for your help on the docs <3

potato4d:
OK!

kwkbtr:
I think this change should be incorporated into the document.
As commented in the issue referred in the above comment (nuxt/nuxt.js#5413):

> Calling and awaiting on `ready()` after the constructor is _still_ recommended.

users should call `ready()` explicitly, but currently there is no mention of `ready()` in the document (except [a brief note in release notes](https://nuxtjs.org/guide/release-notes#programmatic-api-middleware)).
I spent a whole day trying to find the cause of silent 404 errors from Express (the same as in nuxt/nuxt.js#5408).
It would have been a great help if `ready()` was properly documented.

manniL:
@kwkbtr 	Feel free to add it to the docs :relaxed:

PR #1293, opened 2019-04-02
Follow 2.5 release notes.

vlajos:
Follow 2.5 upgrade notes from here: https://nuxtjs.org/guide/release-notes/#programmatic-api-middleware

### DIFF
--- a/en/api/nuxt.md
+++ b/en/api/nuxt.md
@@ -26,6 +26,7 @@ const config = require('./nuxt.config.js')
 
 // Create a new Nuxt instance
 const nuxt = new Nuxt(config)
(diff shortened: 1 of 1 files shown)


en/api/nuxt.md
```diff
@@ -26,6 +26,7 @@ const config = require('./nuxt.config.js')
 
 // Create a new Nuxt instance
 const nuxt = new Nuxt(config)
+await nuxt.ready()
```
pi0: await is not possible on top level code. this requirement has been also removed until nuxt3

vlajos: I see. Thanks.

PR #1296, opened 2019-04-06
nuxt.config.js does not accept import statement

floflock:


manniL:
Nuxt's config **does support ESM syntax**  🤔 

see https://twitter.com/TheAlexLichter/status/1098607134282006541

floflock:
@manniL should be, but have a look?
image

manniL:
If you use express or another `server.js` this doesn't apply that easily as the "entrypoint" doesn't understand ESM syntax. Use `node -r esm server.js` to apply.

floflock:
@manniL thanks. Guess you mean uncompiled ```server/index.js```.

PR #1297, opened 2019-04-07
Add missing HEAD_ATTRS placeholder

drubb:
The HEAD_ATTRS placeholder wasn't documented, I've added  it to reflect the current structure of app.template.html


manniL:
Thanks! Could you change it for the other languages as well? 🤔

drubb:
Ah, didn't notice there are other languages :wink:
I've updated the patch accordingly.

manniL:
Thanks! :+1:

PR #1298, opened 2019-04-07
Update TS Linter Config Docs to avoid false positives with Types only usage

jefer590:
### What does this PR do?
:memo: - Update TS Linter Config Docs to avoid false positives with Types only usage

### Example

When I used the ESLint Code described in the docs, everything went well with the exception of the `no-unused-var` rules that throws a false-positive with types usage only for Typings:

```ts
// This line is throwing `no-unused-vars` linter error
import MyInterface from '~/interfaces/MyInterface'

export default mapper(something: any): MyInterface {
...
}
```

With that config in the docs, users will not have problems with this problem.


kevinmarrec:
@jefer590  I also had this issue, and fell in some Github issues that was talking about this `false positives` behaviors. But if i'm right with your change, it makes a rule toggling an error which is already handled, so how it could fix it then ? Is it like a hidden magic trick ? :P

PR #1310, opened 2019-04-10
Fix small issue with the generate:routeCreated signature.

fillefilip8:
`generate:routeCreated` signature in the docs is `route, path, errors` but in reality, it's `{route, path, errors}`


manniL:
Thanks! :+1:

PR #1313, opened 2019-04-13
Fix link to Nuxt mode

MaxGfeller:
Hi everyone!

I was just going through the docs and found this link that was not working (anymore?). Now i'm not sure if they link i added is the correct one either, so feel free to edit :)


Atinux:
Thanks <3

PR #1315, opened 2019-04-16
zh: update chinese nuxt-render.md to fit new version

melan0802:
Seems like Nuxt updated nuxt.render and nuxt.build usages as well as english doc, but chinese doc haven't updated yet.


melan0802:
No authorization, is that a right status?

Issue #1323, opened 2019-04-17
TypeScript setup broken in newest version


khrome83:
### Version

[2.5.x](https://github.com/docs/releases/tag/2.5.x)

### Reproduction link

[https://nuxtjs.org/guide/typescript/](https://nuxtjs.org/guide/typescript/)

### Steps to reproduce

Follow Installation Guide Here - https://nuxtjs.org/guide/typescript/

### What is expected ?

TypeScript works as expected...

### What is actually happening?

The pages/feed.vue file throws a TypeScript error. 
![Screen Shot 2019-04-17 at 6.51.05 AM.png](https://cmty.s3.eu-west-1.amazonaws.com/74384145/1b9118b0-6107-11e9-bd09-e59dc5c6046e.png)

### Additional comments?

The code is copied and pasted exactly. After tons of research, I can't find a workaround with the exception of setting up `*.d.ts` files and overriding the behavior or turning off `StrictFunctions`. I am new to typescript​ and was really relying on this guide to integrate with it.

<div align="right"><sub><em>This bug report is available on <a href="https://cmty.app/nuxt">Nuxt</a> community (<a href="https://cmty.app/nuxt/docs/issues/c192">#c192</a>)</em></sub></div>

khrome83:
The community TypeScript template also uses `nuxt-property-decorator`. I don't know if that is preferred over `vue-property-decorator`.

khrome83:
Using `nuxt-property-decorator` page as the example, I created this, which does not throw any errors - 

```
<template>
  <section>
    <h1 class="header">Nuxt TypeScript Starter</h1>
    <div class="cards">
      <Card v-for="person in people" :key="person.id" :person="person"></Card>
    </div>
  </section>
</template>

<script lang="ts">
import { Component, Vue } from "nuxt-property-decorator";
import { State } from "vuex-class";
import { Person } from "~/types";
import Card from "~/components/Card.vue";

@Component({
  components: {
    Card
  }
})
export default class extends Vue {
  @State people!: Person;
}
</script>

<style scoped>
.header {
  font-family: "Segoe UI", Tahoma, Geneva, Verdana, sans-serif;
}

.cards {
  display: flex;
  flex-wrap: wrap;
}
</style>
```
I am still knew to typescript,  and just getting back into Nuxt/Vue, so let me know if this is the right direction. If it is, I can make a PR for the doc and update it for you.

kevinmarrec:
@khrome83

[...]

PR #1324, opened 2019-04-18
docs(ts): fix missing `interface` keyword in snippet

khrome83:
Contributes to fixing #1323. Specifically the missing interface keyword.


kevinmarrec:
Thanks @khrome83 !

PR #1327, opened 2019-04-20
all: update guide/index.md

nh0627:
I fixed the term, "ES6/ES7 Transpilation", to "ES2015+ Transpilation" on the guide page. This is because the current Nuxt.js is supporting the latest Javascript😊


manniL:
Many thanks! :+1:

PR #1329, opened 2019-04-23
fr: Update classic mode informations

Assitan:
The information for the Classic mode is not the same as the English version (deprecated in Nuxt 3)


Atinux:
Merci @Assitan :)

PR #1330, opened 2019-04-24
Remove  --project <project-id>

madrus:
The flag `--project` seems not to be supported [any more]. Run `gcloud help app deploy` for more details.

This change in the deployment command is tested on my personal project.

### DIFF
--- a/en/faq/appengine-deployment.md
+++ b/en/faq/appengine-deployment.md
@@ -48,7 +48,7 @@ Now build your app with `npm run build`.
 At this point, your app is ready to be uploaded to Google App Engine. Now just run the following command:
 
 ```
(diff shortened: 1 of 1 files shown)


manniL:
Hey 👋 
Is there any "equivalent"?

madrus:
I have studied the deployment and tested it a bit more:

1. The `gcloud app deploy app.yaml` command is working fine with the default project.
2. However, the key `--project` is also working and can be used to deploy a non-default website. So, my PR is not exactly correct.
3. I suggest to replace `--project <project-id>` with `--project [project-id]` because otherwise the project-id does not print:
project-key

manniL:
@madrus 
Sounds good. Would you change the file then? ☺️

PR #1338, opened 2019-04-27
en: Add TypeScript configuration Example for Express

AustinDavies:


manniL:
Hey @AustinDavies :wave: 

Thanks for the PR! 

I'd like to see some more feedback regarding the proposed changes as it incorporates programmatic usage. cc @pi0 @Atinux @clarkdo

AustinDavies:
> Hey @AustinDavies 👋
> 
> Thanks for the PR!
> 
> I'd like to see some more feedback regarding the proposed changes as it incorporates programmatic usage. cc @pi0 @Atinux @clarkdo

No problem. I expected this to have some amount of back-and-forth before agreeing on a solution XD. Anyway, I know this imposes project style, and perhaps, is even outside the scope of Nuxt itself anyway, however, I found it hard to use Nuxt in 'universal' mode using TypeScript by plainly following the docs; the server framework will error out obviously due the the nuxt.config being an .ts file instead of a transpiled .js file. So, _some_ amount of documentation seemed warrented from the naive user wanting to use Nuxt but not necessarily knowing the in's-and-out's of how to setup a node server in a fully TypeScript environment.

en/guide/typescript.md
```diff
@@ -131,6 +131,132 @@ export default class FeedPage extends Vue {
...
+async function start() {
+  // Init Nuxt.js
+  const nuxt = new Nuxt(config)
+
+  const { host, port } = nuxt.options.server
+
+  // Build only in dev mode
+  if (config.dev) {
+    const builder = new Builder(nuxt)
+    await builder.build()
```
tstoppe89: Why are you awaiting this step, in your medium article you didn't - and as you mentioned in your article: the server will be waiting for this and there would be no loading bars in your browser. I've also tested this behavior.

en/guide/typescript.md
```diff
@@ -131,6 +131,132 @@ export default class FeedPage extends Vue {
...
+...
+```
+
+<div class="Alert Alert--gray">
+
+`node` only supports the "commonjs" module option in the tsconfig.json file.
+
+</div>
+
+If you are NOT using the "commonjs" module option inside your default tsconfig.json file you have two options:
```
tstoppe89: NOT using the "commonjs" module should be the default, as the current documentation says this:
"Notice that es2018 target is needed to be able to use Optional Chaining and Nullish Coalescing, as esnext target doesn't seem to support these features for now."
here: https://typescript.nuxtjs.org/guide/setup.html#configuration

so I guess it would be easy for the reader of your documentation part, that you should assume that everybody has not the commonjs module option. (but all in all - that what you are writing is not wrong, but could be improved a bit)

en/guide/typescript.md
```diff
@@ -131,6 +131,132 @@ export default class FeedPage extends Vue {
...
+```json
+/* server/tsconfig.json */
+/* this is a basic tsconfig.json file. You can add to/remove and modify everything except the module option */
+{
+  "compilerOptions": {
+    "target": "es5",
+    "module": "commonjs",
+    "noEmit": true,
+
+    "strict": true,
```
tstoppe89: since strict is set here, there is also the need to add
`"noImplicitAny": false`
because otherwise the consula import would throw an error for that (that was the problem I had, when following your documentation)

en/guide/typescript.md
pi0: Even for dev mode, there are many async hooks which if any failing we need to properly catch exceptions and also ensure code is loaded. Promise will be resolved as soon as watcher starts so this usage is OK. https://github.com/nuxt/nuxt.js/blob/dev/packages/webpack/src/builder.js#L139

kevinmarrec:
@pi0 What's the deal here ? porting some new TypeScript instructions around programmatic usage with new specs on TS docs ?

debs-obrien:

PR #1344, opened 2019-04-29
ja: fix textlint errors

inouetakuya:


inouetakuya:
I will self-merge this PR because it is just updating formats. 😀

PR #1369, opened 2019-05-10
docs(commands): Update commands.md

ekeyser:
Order of mention should match order of description.
SSR needed to be defined or more clearly explained. There was no clear indication what SSR meant, what it stands for and for which explanation was relevant.


Atinux:
Thanks @ekeyser

PR #1384, opened 2019-05-13
fix: correct context.app type per (#5662)

rchl:
Adjust documentation to match PR https://github.com/nuxt/nuxt.js/pull/5701

Also adjusted some spaces to makes table align in ascii mode


Atinux:
Thanks @rchl

PR #1386, opened 2019-05-14
fr: fix <router-link> docs URL

louistiti:
Fix two dead links for the French translation of the [`<nuxt-link>` page](https://fr.nuxtjs.org/api/components-nuxt-link/).


manniL:
Thanks

PR #1392, opened 2019-05-16
Add documentation for key property

utatti:
Related: https://github.com/nuxt/nuxt.js/pull/5516

There is an undocumented `key` property for page components.

I've added the documentation for EN, KO and JA.

EN and JA are not my mother tongue, so please feel free to review and comment. Will follow up quickly.

cc: @inouetakuya @potato4d


manniL:
Thanks!

PR #1393, opened 2019-05-17
en, de, zh: Fix ignore page path

noliaki:
Fix ignore page path.


manniL:

PR #1395, opened 2019-05-18
doc: remove 'watchQuery' section from 'nuxtChildKey' doc

utatti:
Related: https://github.com/nuxt/nuxt.js/pull/5757

With the pull request above, `watchQuery` no more affects `nuxtChildKey`.

### DIFF
--- a/de/api/components-nuxt.md
+++ b/de/api/components-nuxt.md
@@ -49,10 +49,6 @@ There are 3 ways to handle internal `key` prop of `<router-view/>`.
   }
   ```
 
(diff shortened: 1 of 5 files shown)


manniL:
Will merge when https://github.com/nuxt/nuxt.js/pull/5757 is merged 👍

ThomasR:
* It looks like only half the section was removed.
* In Chinese, the section is still present
* The part "There are 3 ways to handle internal key" shoud be adjusted too (3 -> 2)

PR #1396, opened 2019-05-18
all: escape '<' and '>' in document title

utatti:
`<` and `>` are not shown in document titles.

e.g. https://ja.nuxtjs.org/api/components-nuxt


manniL:

PR #1401, opened 2019-05-18
Link to webpack code splitting is broken on filenames entry

userquin:
https://webpack.js.org/guides/code-splitting-libraries/ is broken on filename entry: replace it with https://webpack.js.org/guides/code-splitting/

### DIFF
--- a/configuration-build.md
+++ b/configuration-build.md
@@ -246,7 +246,7 @@ export default {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


manniL:
Thanks 🙏

3vilArthas:
You have broken this page. Please, put the file in the right place (to "en/api/" not in the root).
https://nuxtjs.org/api/configuration-build

PR #1417, opened 2019-05-27
zh: Add and fix the nuxt module in the documentation

warriorBrian:
## :bug: Fixed

* fix the nuxt module


warriorBrian:
@clarkdo Add and fix completed :smile:

PR #1419, opened 2019-05-28
fix a typo

dskymd:
property 'keys' to 'key'


manniL:
Good catch, thx :relaxed:

PR #1425, opened 2019-06-03
fix: updates `Sentry Module` link

dnlnrs:


manniL:
Hey @dnlnrs, thanks for the fix! Could you change it in other doc languages (if applicable) as well?

aytdm:
Thank you 😊

PR #1439, opened 2019-06-13
en: `extend-webpack.md` with clearer explanation

codebender828:
Added a more elaborate description and example case of how to use the extend method of the build option for the `nuxt.config.js`. Having struggled with finding out how to use this API, I thought this might help future nuxt newbies. Happy to take any advice and make some adjustments! 🚀


manniL:
Many thanks :+1:

codebender828:
Many thanks to you!

PR #1441, opened 2019-06-16
en: add  language identifier to code blocks

aytdm:
ref: https://nuxtjs.org/faq/deployment-azure-portal

To display the code, I added language identifier.


manniL:
Thanks :relaxed:

PR #1455, opened 2019-06-23
fix: use backquotation for the keywords in "API: The Context" chapter

koka831:
some of the keywords in the section does not wrapped in back quotations correctly so that it shows like below; 'middleware', 'modules', and 'nuxtServerInit`.
Screenshot from 2019-06-24 01-52-47

This PR fixes it in some languages which have `api/context.md`.


manniL:
Thanks!

PR #1456, opened 2019-06-23
docs: configuration-generate

debs-obrien:
Sometimes it's not obvious what the generate property is and how to modify it especially if you are not generating dynamic pages therefore don't have a generated property in your nuxt.config.js and one isn't there by default. Adding an example might make things a bit clearer especially for the subFolders option which I think is one that might be used a lot. We could of course add the example for each one to make it even clearer but explaining what other options you could have for example the fallback default is 200, what else would you put or are these things that nobody will change therefore we are just telling people that they are there. It would be great to have a better understanding of some of these generate features.


manniL:
Thank you ☺️

PR #1458, opened 2019-06-26
en: Update development-tools.md

robertoandres24:
Its no longer necessary define babel configs because conflicts with the Ava have by default, and cause some conflicts
In addition, if you write your tests inside a test folder, you need to define it in the package.json, ava object, with the `files` key
https://github.com/avajs/ava#latest-javascript-support


manniL:
Thanks!

PR #1459, opened 2019-06-26
 en: update description for new render:csp option

williamchong007:
reflect changes in https://github.com/nuxt/nuxt.js/pull/5975


manniL:
Thanks!

PR #1468, opened 2019-07-01
Change ssr to mode on plugins guide

JulianMar:


manniL:
Thanks for the addition!
Would you add sth. about mode `server` as well? ☺️

PR #1469, opened 2019-07-02
[Doc FR] All file now aligned with EN version to continue translation

Haeresis:
In order to continue translation, I revised all content.

New page and updated page not already translated are now an FR message that invite the user to participate for FR translation.

All homepage file are now present to start the french translation.

You can merge right now @Atinux !


Atinux:
Wow merci mille fois @Haeresis

PR #1475, opened 2019-07-06
fix: update default scrollBehavior

CurtisBelt:
Notice the documentation here was outdated. Used this as reference: https://raw.githubusercontent.com/nuxt/nuxt.js/918769ee9b2e73c3b8b67a434322235df995cedc/packages/vue-app/template/router.js

I went ahead and overwrote other languages even though the code comments are in English, since I assume when it comes to code snippets that accuracy is more important than being fully translated?


manniL:
Thanks!

PR #1481, opened 2019-07-12
Provide deployment doc for serverMiddleware

garrensweet:
There was no previous indication that serverMiddleware had to be deployed to production and copied anywhere in order to function correctly. This should help users identify that this is a requirement as of now.

### DIFF
--- a/en/api/configuration-servermiddleware.md
+++ b/en/api/configuration-servermiddleware.md
@@ -84,3 +84,8 @@ serverMiddleware: [
     '~/api/logger'
 ]
 ```
+
+## Deploying Server Middleware
+
+When deploying server middleware it is a requirement that the middleware is also copied over during the build proces. It should reside in the `dist/server` directory.
+


en/api/configuration-servermiddleware.md
```diff
@@ -84,3 +84,8 @@ serverMiddleware: [
     '~/api/logger'
 ]
 ```
+
+## Deploying Server Middleware
+
+When deploying server middleware it is a requirement that the middleware is also copied over during the build proces. It should reside in the `dist/server` directory.
```
manniL:
```suggestion
When deploying server middleware it is a requirement that the middleware is also copied over during the build process. It should reside in the `dist/server` directory.
```

manniL:
Thank you! ☺️

Atinux:
Well actually they should not be copied, for some hosting like Now, our builder is actually doing the work of copying the `serverMiddleware`.

PR #1497, opened 2019-07-24
fr: fix some links

Alvai:
Vuex link redirected to `https://ssr.vuejs.org/fr` which is not the correct library

Vue Server Renderer link redirected to `https://ssr.vuejs.org/fr` which is a 404 page because `ssr.vuejs.org` doesn't have french translation yet


manniL:
Merci!

PR #1499, opened 2019-07-28
Add display=swap paramater

lamalamaMark:
&display=swap is added to the default url in Google Fonts


manniL:
Sure, better perf by default doesn't hurt :speak_no_evil:

PR #1500, opened 2019-07-28
Publish directory when deploying in SPA mode should be .nuxt/dist

gjigandi:
The deploy in Netlify failed when setting the publishing directory 'dist'.
Changing to '.nuxt/dist' seems to work.


Atinux:
Actually using `nuxt build` with `mode: 'spa'` will generate a `dist/` directory :)

PR #1518, opened 2019-08-10
move typescript docs to dedicated page

pi0:


manniL:
Thanks Pooya :raised_hands:

PR #1539, opened 2019-08-21
ru: fix error about excluding vue-server-renderer

alexklyuev:
Existing translation of "excluded when using mode: 'spa'"  as "только при использовании mode: 'spa'" is quite the opposite of its actual meaning, this pr changes it to the right meaning.


debs-obrien:

PR #1541, opened 2019-08-23
en: improve generate.fallback

manniL:


en/api/configuration-generate.md
```diff
@@ -91,9 +91,11 @@ export default {
 }
 ```
 
-The path to the SPA fallback. This file can be used when doing deploys of generated sites to static hosting. It falls back to `mode: 'spa'` when a route isn't generated.
+The path to the fallback HTML file. It should be set as the error page, so that also unknown routes are rendered via Nuxt.
+If set to `true`, it will render as `404.html`.
```
pimlie: To remove all ambiguity i would suggest to explicitly list all the options.

```suggestion
If unset or set to a falsy value, the filename of the fallback HTML file will be `200.html`. If set to `true`, the file will be rendered as `404.html`. In all other cases the filename you provided will be used.
```

manniL: I've reworded it a little. WDYT? :)

debs-obrien:
Looking much better however As per chat with @manniL I recommend we explain it better by adding the following text or similar with perhaps an example underneath of the fallback:true/false etc
1. SPA with all pages not generated -> 200.html, as it's the only file necessary as no other routes are generated
2. Statically generated pages -> 404.html for error pages and those covered by excludes
However, you can also go always with #2

manniL:
Closed in favor in https://github.com/nuxt/docs/pull/1552

PR #1551, opened 2019-08-24
Update netlify-deployment.md

debs-obrien:
modify Netlify example for spa fallback


debs-obrien:
@pimlie @manniL please check, thanks

PR #1552, opened 2019-08-24
en: improve generate.fallback

debs-obrien:
Better explanation and examples for 404/200 fallback @manniL @pimlie please check to see if it is too much or is missing anything


manniL:
Let's get this in sync with #1541

en/api/configuration-generate.md
```diff
@@ -81,7 +81,7 @@ export default {
 ## fallback
 
 - Type: `String` or `Boolean`
-- Default: `'200.html'`
+- Default: `200.html` if left unset or false and `404.html` if set to true
```
manniL: The outcome for different values is explained further down, so I'd suggest to keep it as is :relaxed: 	

```suggestion
- Default: `200.html`
```

en/api/configuration-generate.md
```diff
@@ -91,9 +91,30 @@ export default {
 }
 ```
 
-The path to the SPA fallback. This file can be used when doing deploys of generated sites to static hosting. It falls back to `mode: 'spa'` when a route isn't generated.
+The path to the fallback HTML file. It should be set as the error page, so that also unknown routes are rendered via Nuxt.
+If unset or set to a falsy value, the name of the fallback HTML file will be `200.html`. If set to `true`, the filename will be `404.html`. If you provide a string as a value, it will be used instead.
 
-_Note: this option could be useful using [Netlify](https://netlify.com) or any static hosting using HTML fallbacks._
+If running a SPA with all pages not generated then it is best to fallback to 200.html, as it's the only file necessary as no other routes are generated.
```
manniL:
```suggestion
When running an SPA then it is more idiomatic to use a `200.html`, as it's the only file necessary as no other routes are generated.
```

en/api/configuration-generate.md
```diff
@@ -91,9 +91,30 @@ export default {
...
+
+```js
+fallback: false
+```
+
+If working with statically generated pages then it is best to fallback to 404.html for error pages and for those covered by [excludes](https://nuxtjs.org/api/configuration-generate#exclude) (the files that you do not want generated as static pages).
```
manniL:
```suggestion
If working with statically generated pages then it is recommended to use a `404.html` for error pages and for those covered by [excludes](https://nuxtjs.org/api/configuration-generate#exclude) (the files that you do not want generated as static pages).
```

en/api/configuration-generate.md
```diff
@@ -91,9 +91,30 @@ export default {
...
+
+However, Nuxt allows you to configure any page you like so if you don't want to use the 200.html or 404.html you can add a string and then you just have to make sure you redirect to that page instead. This is of course not necessary and is best to redirect to 200.html/404.html
+
+```js
+fallback: 'fallbackPage.html' 
+```
+
+ *Note: Multiple services (e.g. Netlify) detect a `404.html` automatically.*
+ 
+ *Note: If running on apache you will need to create a .htaccess file with a redirect of 404 to 404.html.*
```
manniL: We can't add comments for all web servers (nginx, apache, etc.), so I'd either leave it out or make it more general. Maybe even merging it with 115

en/api/configuration-generate.md
```diff
@@ -111,7 +132,7 @@ Interval between two render cycles to avoid flooding a potential API with API ca
 
 - Type: `Array`
 
-[Dynamic routes](/guide/routing#dynamic-routes) are ignored by the `generate` command.
+[Dynamic routes](/guide/routing#dynamic-routes) are ignored by the `generate` command (npm run generate). Nuxt does not know what these routes will be so it can't generate them.
```
manniL:
```suggestion
[Dynamic routes](/guide/routing#dynamic-routes) are ignored by the `generate` command (yarn generate). Nuxt does not know what these routes will be so it can't generate them.
```

en/api/configuration-generate.md
```diff
@@ -124,7 +145,7 @@ Example:
 
 Only the route `/` will be generated by Nuxt.js.
 
-If you want Nuxt.js to generate routes with dynamic params, you need to set an array of dynamic routes.
+If you want Nuxt.js to generate routes with dynamic params, you need to set an array of dynamic routes inside the `generate` property.
```
manniL:
```suggestion
If you want Nuxt.js to generate routes with dynamic params, you need to set the `generate.routes` property to an array of dynamic routes.
```

en/api/configuration-generate.md
debs-obrien: resolved

debs-obrien:
@manniL changes made as requested. please check

en/api/configuration-generate.md
```diff
@@ -91,10 +91,29 @@ export default {
...
+fallback: true
+```
+
+However, Nuxt allows you to configure any page you like so if you don't want to use the `200.html` or `404.html` you can add a string and then you just have to make sure you redirect to that page instead. This is of course not necessary and is best to redirect to `200.html`/`404.html`.
+
+```js
+fallback: 'fallbackPage.html' 
+```
+
+ *Note: Multiple services (e.g. Netlify) detect a `404.html` automatically. For those that don't you will need to manually create a redirect file that redirects `404` to `404.html` *
```
manniL:
```suggestion
 *Note: Multiple services (e.g. Netlify) detect a `404.html` automatically. If you configure your webserver on your own, please consult it's documentation to find out how to set up an error page (and set it to the `404.html` file) * 
```

PR #1553, opened 2019-08-25
ja: add faq/using-nuxt-with-cloudflare

aytdm:
PTAL, when you're available.
resolves https://github.com/vuejs-jp/ja.docs.nuxtjs/issues/234, resolves https://github.com/vuejs-jp/ja.docs.nuxtjs/issues/235


aytdm:
Thanks ❤️

PR #1557, opened 2019-08-28
en: update firebase custom 404.html handling

skl:
Firebase Hosting now automatically supports `404.html` with a default 404 status code without having to modify `firebase.json` anymore. Using the previous config from these docs actually resulted in a response code of 200 for Page Not Found errors.


manniL:
Thank you!

PR #1566, opened 2019-09-04
en: fix typo

konovalov-nk:
Many people adhere to a belief that you use the article “a” before words that begin with consonants and “an” before words that begin with vowels. But that isn’t the rule.

The real rule is this: You use the article “a” before words that start with a consonant sound and “an” before words that start with a vowel sound. For example, "He has a unique point of view on the subject and talked about it for an hour". The “u” in “unique” makes the “Y” sound—a consonant sound—therefore you use “a” as your article, while the “h” in “hour” sounds like it starts with “ow” — a vowel sound.

(c) Brian A. Klems, https://www.writersdigest.com/online-editor/a-before-consonants-and-an-before-vowels-is-not-the-rule

### DIFF
--- a/en/api/configuration-head.md
+++ b/en/api/configuration-head.md
@@ -31,6 +31,6 @@ You can also use `head` in your components and access to the component data thro
 
 <div class="Alert Alert--teal">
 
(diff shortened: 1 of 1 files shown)


manniL:
Thanks! ☺️

PR #1577, opened 2019-09-09
docs: find app.html in src dir instead of root dir

iliyaZelenko:
Re https://github.com/nuxt/nuxt.js/pull/6374


manniL:
Thanks mate!

pi0:
Thanks!

PR #1581, opened 2019-09-13
i.e. to e.g.

ideabrian:
update latin acronym


ideabrian:
i.e. means that is, e.g. is for example

manniL:
Thanks!

PR #1584, opened 2019-09-14
Update installation.md

ideabrian:
Some wording felt more clear to me by changing out "the" with "a" when you're talking about something they have to create. When you say "the" I'm asking myself -- where is it?


Atinux:
Thanks @ideabrian

PR #1586, opened 2019-09-17
The paragraph for Async Modules was out of place

mrjimenez:
The paragraph for "Async Modules" was out of place, it was mixing information with "Build-Only Modules".


debs-obrien:

PR #1589, opened 2019-09-21
Updated code snippet to user $axios.$get.

DominusKelvin:
To use axios inside of the fetch method, user should use $axios.$get instead of axios.get because the latter doesn't work when using Axios module.

### DIFF
--- a/en/api/pages-fetch.md
+++ b/en/api/pages-fetch.md
@@ -31,7 +31,7 @@ Example of `pages/index.vue`:
 <script>
 export default {
   fetch ({ store, params }) {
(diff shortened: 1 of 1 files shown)


manniL:
In the docs we assume that no external module is used, as users might not do. I'd stick to the default axios syntax *or* revamp the parts of the docs to highlight that the axios module is used.

Besides that, adding `$` signs only wouldn't be enough :relaxed: 
`const { data } = axios.get()` vs. `const { data } = $axios.get()` vs. `const data = $axios.$get()`

Thanks anyway! :relaxed:

PR #1601, opened 2019-09-28
the file import url is not right

fillfive:
`import axios from '~plugins/axios'` 改为 `import axios from '~/plugins/axios'`


manniL:
Both works :relaxed:

fillfive:
image
image
it error when I use this.

fillfive:
@manniL

manniL:
Interesting. This *should* work AFAIK.

Thanks for the info.

fillfive:
Thank you. the nuxt docs help me a lot. Thanks for your efforts.

manniL:
That's great to hear! Appreciate the feedback :relaxed: 

Thanks for your contribution mate :pray:

PR #1605, opened 2019-10-02
Noting that serverMiddleware can't access nuxt env variables

denilsonsa:
Based on this answer: https://github.com/nuxt/nuxt.js/issues/2800#issuecomment-414041637

(It took me quite some time to figure out why it wasn't working on my case; so I'm documenting what's the expected behavior.)


Atinux:
Thank you, nice point :)

PR #1607, opened 2019-10-02
Changing the search text to indicate the availability of focus

thiyagu-7:
This is a follow up PR from https://github.com/nuxt/nuxtjs.org/issues/233 for changing the search box's placeholder text.

Before:
image

With the placeholder text change:
image


Atinux:
Perfect 💚

PR #1640, opened 2019-10-12
en: add head function hint

amdp:
Added intro in order to suggest the use of head function.


en/faq/external-resources.md
```diff
@@ -5,6 +5,10 @@ description: How to use external resources with Nuxt.js?
 
 ## Global Settings
 
+You can include your external resources in the head object or function.
+The following example is with an object, the second one is with a function. 
+If you encounter problems you may would like to switch to the head function.
```
manniL: Especially when you want to use values coming from the Vue component (so all "non-static" content), a function is the only way to go (and to ensure dependencies are tracked and reactivity works).

It'd be nice if you could mention that somewhere :relaxed:

amdp:
In my case the object seems to work better. I will try to make a sandbox
and check.

Thank you,
Alessandro Merletti De Palo
https://amdp.me

Il dom 13 ott 2019, 10:26 Alexander Lichter <notifications@github.com> ha
scritto:

en/faq/external-resources.md
```diff
@@ -5,6 +5,10 @@ description: How to use external resources with Nuxt.js?
 
 ## Global Settings
 
+You can include your external resources in the head object or function.
+The following example is with an object, the second one is with a function. 
+If you encounter problems you may would like to switch to the head function: especially when you want to use values coming from the Vue component (so all "non-static" content), a function is the only way to go (and to ensure dependencies are tracked and reactivity works).
```
manniL:
```suggestion
If you want to use values from your Vue component like computed properties or data, you can also use `head` as a function, returning the final head object.
```

manniL:
thanks!

PR #1644, opened 2019-10-14
Remove unused key and value from lang.json

aytdm:
Organized `lang.json` as part of the translation. Specifically, I removed keys and values ​​that are not imported by [nuxtjs.org](https://github.com/nuxt/nuxtjs.org). I have confirmed that there are no problems, but please confirm it just in case.

Ref: https://github.com/nuxt/nuxtjs.org/issues/263

### DIFF
--- a/de/lang.json
+++ b/de/lang.json
@@ -2,73 +2,28 @@
   "iso": "en",
   "docVersion": "1.4.0",
   "links": {
-    "api": "API",
-    "blog": "Blog",
-    "chat": "Chat",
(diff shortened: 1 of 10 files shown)


Atinux:
Thank you very much @aytdm 

PS: really nice to meet you again in Japan :)

aytdm:
@Atinux Thank you for your confirmation. I was gladded to meet you the other day :) I'm looking forward to seeing you again! (Hope I'll be able to speak more fluently by the time we meet again.)

PR #1661, opened 2019-10-18
Document that watchQuery can be a function

### LINKED ISSUE #1631: Document that watchQuery can be a function
Document https://github.com/nuxt/nuxt.js/pull/6297

studnitz:
Also, add an example using the function API.

This closes #1631


manniL:
Thanks!

PR #1686, opened 2019-10-24
fr: update views.md first section title

antoineneff:
I'm not sure about this one but something should be done as the menu is using `Document` and the views.md section title related is `modèle de l'application` so the link does not work.

I don't think this term is really used in french and `Document` is simple enough.


Atinux:
Merci !

Issue #1628, opened 2019-10-08
Document extendPlugins


manniL:
Document https://github.com/nuxt/nuxt.js/pull/6285

starikovm:
On it

PR #1697, opened 2019-10-26
fix nuxt/docs#1628

starikovm:
- fix nuxt/docs#1628, adding extendPlugins docs
- fix internals builder `build:prepare` hook to `builder:prepare`

### DIFF
--- a/en/api/internals-builder.md
+++ b/en/api/internals-builder.md
@@ -19,7 +19,8 @@ this.nuxt.hook('build:done', (builder) => {
 Hook                 | Arguments                                  | When
 ---------------------|--------------------------------------------|----------------------------------------------------------------------------------------------------------------------------------------
 `build:before`       | (nuxt, buildOptions)                       | Before Nuxt build started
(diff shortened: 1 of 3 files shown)


en/api/internals-builder.md
```diff
@@ -19,7 +19,8 @@ this.nuxt.hook('build:done', (builder) => {
 Hook                 | Arguments                                  | When
 ---------------------|--------------------------------------------|----------------------------------------------------------------------------------------------------------------------------------------
 `build:before`       | (nuxt, buildOptions)                       | Before Nuxt build started
-`build:prepared`     | (nuxt, buildOptions)                       | The build directories have been created
+`builder:prepared`     | (nuxt, buildOptions)                       | The build directories have been created
```
starikovm: I saw that actual hook [here](https://github.com/nuxt/nuxt.js/blob/dev/packages/builder/src/builder.js#L155) is `builder:prepared` instead of `build:prepared`

Issue #1295, opened 2019-04-05
Improve build.extends explanation / examples


manniL:
https://nuxtjs.org/faq/extend-webpack/ isn't very helpful.

debs-obrien:
Thanks for your contributions. We are now closing this repo as docs have now moved to nuxtjs.org using content module. Please continue your contributions on that site especially focusing on the guides folder which is the new docs. Thanks

https://github.com/nuxt/nuxtjs.org/blob/master/README.md

PR #1699, opened 2019-10-26
add docs to faq/extend-webpack (nuxt/docs#1295)

starikovm:
title speaks for itself


husayt:
This is useful. Thanks

PR #1708, opened 2019-10-29
Removed improper use of the adjective 'simply'

kdekooter:
I have removed the word "simply" in two places. The first instance is plain incorrect: this document is far from complete. "Simply" exporting something (how, from where, synchronous, asynchronous, in the deprecated index.js?) does not work at all. And the second instance is plain unnecessary.


Atinux:
Thanks @kdekooter

PR #1713, opened 2019-10-30
Update webpack-plugins.md

gangsthub:
- Provide more context.
- Explain the why `ProvidePlugin` is in the docs and what it does.
- And slight difference with `build.extend` where you can also pass webpack plugins.


Atinux:
Thank you @gangsthub

PR #1728, opened 2019-11-06
Fix a typo in the installation section

afzalsayed96:


aytdm:
Thank you for your PR. But I close this PR because [Create nuxt app#Usage](https://github.com/nuxt/create-nuxt-app#usage) is `yarn create nuxt-app <my-project>` and this description is correct.

PR #1739, opened 2019-11-14
Added postcss plugins & nuxt-tailwindcss setup section

samuells:
Added section where is explain how to apply the postcss plugin on the configuration from nuxt-tailwindcss


Atinux:
Awesome thanks @samuells

PR #1743, opened 2019-11-18
zh: migrate ts to new doc

zgayjjf:
As nuxt's support for typescript has been moved to standalone sites, there is no need for a legacy document (which confused me a while).

Wrong/Legacy docs: [https://zh.nuxtjs.org/guide/typescript/](https://zh.nuxtjs.org/guide/typescript/)
Right docs: [https://typescript.nuxtjs.org/](https://typescript.nuxtjs.org/)


zgayjjf:
Besides, I noticed that the Chinese docs haven't been updated for a long time. I wonder if I can be the maintainer of Chinese docs? 

If the answer is yes, where should I begin?

zgayjjf:
Any idea?

PR #1752, opened 2019-11-25
Update directory-structure.md

abdurrahmanriyad:
In component asyncdata or fetch either can be used. By mistake, it was written as can't.


Atinux:
Actually you cannot use `asyncData` in `components/`

PR #1771, opened 2019-12-14
CSS/SCSS properties had bad path

sbarre:
The example paths for the SCSS and CSS files should not have a `/` after the `@` character. 

This causes a rather inscrutable runtime error: 
```
Could not compile template /Users/[...]/node_modules/@nuxt/vue-app/template/App.js: Cannot resolve "@/assets/css/main.scss" from "/Users/[...]/assets/css/main.scss"
```


Atinux:
Do you have a reproduction @sbarre ?

sbarre:
@Atinux what do you mean a reproduction?

I pasted the error message I got when trying to build after following the documentation on this page:

https://nuxtjs.org/api/configuration-css

So after a bunch of digging on Stack Overflow, I found out the problem, which was to remove the `/` character, and I submitted a PR.

I apologize, I'm a bit new at this, can you explain in more detail what else you need, and I will try to provide it?

debs-obrien:

PR #1778, opened 2019-12-21
en(routing): add info about accessing parameters

danpastori:
Added a small section to display how to access dynamic route parameters within a local component or page. I added this to clarify any confusion about how to access these parameters.


manniL:

PR #1825, opened 2020-01-29
Change example of Default Meta Tags

venerated:
head is shown as an object instead of a function.


husayt:
@venerated the example as it is is  also correct. Both are valid syntax. See [here](https://nuxtjs.org/api/pages-head#the-head-method)

venerated:
When I used the current syntax in the documentation, it caused an error.

PR #1829, opened 2020-02-04
updated nuxtServerInit verbiage a little.

hecktarzuli:


husayt:
@hecktarzuli  this indeed is more clear. Thanks

PR #1838, opened 2020-02-19
changed nuxt.config.js import statement to require

sugoidesune:


Atinux:
You can use `import` with `nuxt.config.js` actually

PR #1840, opened 2020-02-24
Fix env name for baseURL

fabienoger:
Hello,

I encountered a problem today which allowed me to identify an error in the documentation (I think).

The `baseUrl` env variable is not considered in my `nuxt.config.js` so I tried `baseURL` which works.

[In the docs](https://nuxtjs.org/api/configuration-env/) it's actually `baseUrl` so my PR only change `baseUrl` into `baseURL`.

My version of Nuxt.JS is 2.11.0
Hope my update is good !

Thank you all for your work :)

### DIFF
--- a/de/api/configuration-env.md
+++ b/de/api/configuration-env.md
@@ -14,17 +14,17 @@ Example (`nuxt.config.js`):
 ```js
 module.exports = {
   env: {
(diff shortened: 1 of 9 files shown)


manniL:
Hey 👋🏻

it's not wrong use either version ☺️

PR #1846, opened 2020-02-27
docs(eslint-loader): use cache

iliyaZelenko:
This will fix performance in very large projects. For example, in our project, the `dev` mode (`yarn dev`) was launched without this for **30 minutes**, and after using the cache, for **50 seconds**!!!


debs-obrien:

PR #1859, opened 2020-03-11
Fix google-analytics route error when using Adblock

bdrtsky:
When using Adblock extention (in my case uBlock), this route is broken - https://nuxtjs.org/faq/google-analytics

I suggest to rename it to `/faq/ga` to pass Adblock rules. Or may be even some random value (but `ga` seems to work)


Atinux:
Nice one @bdrtsky 

To avoid breaking the SEO, we also need to add a redirect here: https://github.com/nuxt/nuxtjs.org/blob/master/_redirects

```bash
/google-analytics /ga
```

bdrtsky:
Good point, I'll update it.

PR #1881, opened 2020-03-23
add fetchOnServer/fetchDelay example

bdrtsky:
Currently documentation doesn't provide visual example how to use fetchOnServer/fetchDelay options.


Atinux:
Thank you @bdrtsky

PR #1882, opened 2020-03-23
Corrected a grammatical error in the Homepage

neelansh15:
Change: ~~killing~~ => __kill__


neelansh15:
**Current sentence:** _Instead of switching to another framework and
spending time to get used to it, why not ~~killing~~ two birds with one stone?_
**New Sentence:**  _Instead of switching to another framework and
spending time to get used to it, why not **kill** two birds with one stone?_

PR #1897, opened 2020-04-05
fix: correct description of fetchDelay

tony19:
The current description of `fetchDelay` states:

> set the minimum executing time in milliseconds (to avoid quick flashes)

While it's true that a delay would establish a minimum execution time, that description itself is not particularly descriptive of its purpose. I think a more helpful description would be:

> milliseconds to wait before calling `fetch()` (to avoid rendering flashes)


Atinux:
Actually it does not wait X ms before calling fetch, it just set a minimum execution time, if `fetch` is 120ms, we will wait 80ms more so it is 200ms at the end

tony19:
Oh, I see now. Thanks

PR #1900, opened 2020-04-07
en: don't pin pug dep

mogya:
Pug is now 2.0.4 released. And we have no reason to specify the Pug version anymore.


manniL:
Thanks!

PR #1902, opened 2020-04-09
en: remove useless section of heroku deployment.

mathieutu:
Hi, 
the NPM_CONFIG_PRODUCTION=false is outdated, and not necessary anymore:

>By default, Heroku will install all dependencies listed in package.json under dependencies and devDependencies.
>After running the installation and build steps Heroku will strip out the packages declared under devDependencies before deploying the application.

https://devcenter.heroku.com/articles/nodejs-support#package-installation


manniL:
Thanks!

PR #1910, opened 2020-04-16
doc: add warning for using plugins

SasanFarrokh:
Warning for this issue:

https://github.com/nuxt/nuxt.js/issues/7207


Atinux:
Need to be reviewed a bit more since it is possible but need an extra check to avoid bad memory usage.

PR #1916, opened 2020-04-19
remove first forward slashes in aliases example

bergjs:


Atinux:
Actually it works when the `/`

PR #1919, opened 2020-04-20
added server/index.js config needed for https operation and additiona…

cpachmayr:
…l info on creating/using keys in localhost development environment


cpachmayr:
I'm curious as to why the needed code changes for server.js weren't accepted with the merge?  That's actually the critical part that took me a while to figure out and prompted me to suggest the changes to the docs.

PR #1927, opened 2020-04-22
en: remove unnecessary env variable from GAE deploy

theomessin:
`NODE_ENV` is set to `production` by Google App Engine itself. No need to redefine it.
https://cloud.google.com/appengine/docs/standard/nodejs/runtime#environment_variables


manniL:
thanks!

PR #1943, opened 2020-05-02
en: add clarification around watchQuery and the new fetch hook

andysleigh:
This is to resolve issue [7281](https://github.com/nuxt/nuxt.js/issues/7281)


HunderlineK:
Hi, can we have this rebased and merged? :D Took me an hour to figure out what's going one, thanks!

manniL:
Thanks 👍

PR #1947, opened 2020-05-06
ko: update menu.json of /api

nh0627:


nh0627:
There is an error while linting. Could you fix this @Atinux ?

PR #1972, opened 2020-05-19
en: add deploy docs for Azure Static Web Apps

debs-obrien:
Add docs for Azure Static web apps just announced now at MSBuild. New page and added to menu


en/faq/deployment-azure-static-web-apps.md
```diff
@@ -0,0 +1,111 @@
...
+1. Click the **Sign in with GitHub button**
+2. Select the **Organization** under which you created the repo for your Nuxt.js project. It can also be your GitHub username.
+3. Find the name of the repository you created earlier and select it.
+4. Choose **master** as the branch from the *Branch* dropdown.
+
+![how to add github](https://user-images.githubusercontent.com/13063165/82118359-38ea4100-9777-11ea-9c5e-7ba5c4da708e.png)
+
+### Step 3: **Configure the build process**
+
+There are few things that Azure App Service Static App can assume--things like automatically installing npm modules and running `npm run build`. There are also few you have to be explicit about, like what folder will the static app be copied to after build so the static site can be served from there.
```
manniL, line 64:
```suggestion
There are few things that Azure App Service Static App can assume - things like automatically installing npm modules and running `npm run build`. There are also few you have to be explicit about, like what folder will the static app be copied to after build so the static site can be served from there.
```

PR #1973, opened 2020-05-20
Add a warning about memory leak

pand-app:
Proof of memory & CPU leak : 
(red line is a container restart)
Before, with Vue.use in the exported function of a plugin : 
Screenshot 2020-02-06 at 09 24 39
Screenshot 2020-02-06 at 09 19 58

After, moved Vue.use outside of this exported function : 
Screenshot 2020-02-10 at 11 15 31

### DIFF
--- a/en/guide/plugins.md
+++ b/en/guide/plugins.md
@@ -148,6 +148,10 @@ export default {
 }
 ```
 
+<div class="Alert">
+NB : Don't use a `Vue.use()`, `Vue.component()`, and globally, don't plug anything in Vue **inside** this function, dedicated to Nuxt injection. It will cause memory leak, server-side.
+</div>
+
 ### Combined inject
 
 If you need the function in the `context`, Vue instances and maybe even in the Vuex store, you can use the `inject` function, which is the second parameter of the plugins exported function.


debs-obrien:

debs-obrien:
pr added to new docs - https://github.com/nuxt/nuxtjs.org/pull/505

elcferreira:
Hello Dears, I've developed a nuxt-module, and I'm having a huge problem with memory leak.

image


I saw that I should not have global components on plugins, and I'm not understanding where should I put this code.

I'm using this template here:
https://github.com/nuxt-community/module-template/tree/master/template

Basically, I'm registering the components on `addPlugin`.

```javascript
  // modules.js

  // register the plugins
  const pluginsToSync = [
    'store.js',
    'components.js'
  ]

  for (const pathString of pluginsToSync) {
    this.addPlugin({
      src: resolve(__dirname, pathString),
      fileName: join(namespace, pathString),
      options
    })
  }
```

And on the `components.js`.
```javascript
import Vue from 'vue'
import components from './components/index.js'

// loop through components and register them
for (const name in components) {
  Vue.component(name, components[name])
}
```

Is it wrong? Have someone some example how to register global components on `nuxt-modules`?
I have checked different libraries and they do in this way, but for me it's causing memory leak. 😢 

Thank you very much in advance.

❤️

PR #1980, opened 2020-05-25
fix: Function needs to be injected in context.app

6ones:
To access the injected function in the context, it needs to be explicitly injected into it.
`app.injectedFunction = injectedFunction`
Without doing this, I realized it did not exist in the `context.app` object


manniL:
When using `inject('myInjectedFunction', myInjectedFunction) ` it will exist as `context.app.$myInjectedFunction`.

PR #2013, opened 2020-06-05
en(blog): improve wording in fetch article

manniL:
Makes clear that we can use `this.$nuxt.error` inside `fetch`.


manniL:
WDYT @Krutie?

Krutie:
> WDYT @Krutie?

That's perfect! Thanks.

PR #2014, opened 2020-06-06
en(directory-structure): fix link to `asyncData`

matthaias:
I think this should link to the `asyncData` page but is wrong and was closed with a `]` instead of a `)` accidentally.


manniL:
Thanks!

PR #2091, opened 2020-07-06
Update vuex-store.md

Abdulazeez-Dolapo:
Make the description of nuxtServerInit less confusing.


debs-obrien:
Thanks, we are releasing our new docs this week which should fix all these issues. if however they don't then a pr on the new docs would be great. thanks

PR #2105, opened 2020-07-14
Add way to disable showing error page

mojighahar:


debs-obrien:
thanks, we will look into this as working on releasing new docs soon.

PR #2107, opened 2020-07-15
duplicate words

HomyeeKing:
At the #Creating a previous and next component part,
below the `components/PrevNext.vue` source code:
image


debs-obrien:
thanks, just modified in new docs that will be released next week

PR #2121, opened 2020-07-24
Update github-pages.md

ndragun92:
Added missing explanation on how to call assets right way while deploying app to gh-pages.


debs-obrien:

PR #2122, opened 2020-07-24
Bug fix in code for vuex-store.md

lance-anderson:
`todo` is null and consequently the last todo is removed instead of the specified todo.


debs-obrien:

[new docs](https://github.com/nuxt/nuxtjs.org/blob/master/README.md)

PR #2124, opened 2020-07-27
Improve documentation for context.redirect

unikitty37:
I had to search through the source to find out how to use `query`, which is not ideal. Hopefully this will help the next person in line.


debs-obrien:

PR #2135, opened 2021-10-17
docs(en,ja): add explanation for 'nuxt generate --fail-on-error=[number]'

aewshyae:
add explanation for 'nuxt generate --fail-on-error=[number]' 

languages: en, ja

(now I prepare for PR in nuxt/nuxtjs )

updated: 
https://github.com/nuxt/nuxt.js/pull/9956


aewshyae:
close and then move to https://github.com/nuxt/website-v2

